PR #6, opened 2025-06-05
Add CenterNet model in ONNX format

devisettymahidhar608:
Description
This PR adds support for models such as CenterNet.

### DIFF
--- a/centernet/pytorch/loader.py
+++ b/centernet/pytorch/loader.py
@@ -1,0 +1,60 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""
+CenterNet model loader implementation
+"""
+import torch
+import onnx
+from PIL import Image
+from torchvision import transforms
+import numpy as np
+
+from ...base import ForgeModel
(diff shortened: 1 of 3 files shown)


centernet/pytorch/loader.py
```diff
@@ -0,0 +1,72 @@
...
+
+from ...base import ForgeModel
+from ...tools.utils import get_file
+
+
+class ModelLoader(ForgeModel):
+    """CenterNet model loader implementation."""
+
+    @classmethod
+    def load_model(cls, dtype_override=None):
```
kmabeeTT, line 21: plz remove dtype_override arg and description here since not used

centernet/pytorch/loader.py
```diff
@@ -0,0 +1,72 @@
...
+        """Load and return the CenterNet model instance with default settings.
+
+        Args:
+            dtype_override: Optional torch.dtype to override the model's default dtype.
+                           If not provided, the model will use its default dtype (typically float32).
+
+        Returns:
+            Onnx model: The CenterNet model instance.
+        """
+        # Load model with defaults
```
kmabeeTT, line 31: cool, this is fine, eventually we will add multiple variants to this model and others, and a query function.

PR #17, opened 2025-06-18
Add ONNX Model Variants of CenterNet for OD, HPE, and 3D_BB

devisettymahidhar608:
**Problem description:**
This PR introduces the CenterNet model in ONNX format for multiple tasks, including **Object Detection** (OD), **Human Pose Estimation** (HPE), and **3D Bounding Box** (3D_BB).
 The implementation is based on the official CenterNet [repository](https://github.com/xingyizhou/CenterNet) 
Supported model variants for each task are:

- Object Detection: dla1x, dla2x, resdcn101, resdcn18

- Human Pose Estimation: dla1x, dla3x

- 3D Bounding Box Detection: ddd_3dop, ddd_sub

### DIFF
--- a/centernet/pytorch/loader.py
+++ b/centernet/pytorch/loader.py
@@ -18,36 +18,43 @@ class ModelLoader(ForgeModel):
     """CenterNet model loader implementation."""
 
     @classmethod
-    def load_model(cls):
+    def load_model(cls, **kwargs):
         """Load and return the CenterNet model instance with default settings.
 
         Returns:
             Onnx model: The CenterNet model instance.
         """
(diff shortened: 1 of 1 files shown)


kmabeeTT:
Only suggestion I have is to use `variant_name` instead of `model_name` here, so model_name variable (in tests) can contain more verbose info like `model_name = f"centernet-{variant_name}`.  So this change I would suggest:

```
- model_name = kwargs.get("model_name", "dla1x_od")
+ variant_name = kwargs.get("variant_name", "dla1x_od")
```

PR #23, opened 2025-06-23
Model naming

ashokkumarkannan1:
### Ticket
https://github.com/tenstorrent/tt-forge-fe/issues/2264

### Description 
This PR Includes the ModelArch Enum which can be used by every Frontend


mrakitaTT:
@nvukobratTT Can we please first have discussion about changes like this before merging something that all frontends should use? :)

As mentioned in discussion on slack, I don't see much point in having enum with the names of all models we support, and also we are already tackling this with @kmabeeTT's PR [here](https://github.com/tenstorrent/tt-forge-models/pull/7). I'm open to be convinced we need this, but would just like to have a chance for discussion before we make changes :)

ashokkumarkannan1:
Hi @nvukobratTT,
Can we update the tt-forge-fe test scripts to track these changes?
Let me know if you have any suggestions .

PR #30, opened 2025-06-25
Add post-processing steps for newly added models:set1

meenakshiramanathan1:
This PR adds post-processing class method for newly added models in tt_forge_models repo.

### DIFF
--- a/alexnet/pytorch/loader.py
+++ b/alexnet/pytorch/loader.py
@@ -16,7 +16,7 @@ from ...config import (
 from ...base import ForgeModel
 
 from PIL import Image
-from ...tools.utils import get_file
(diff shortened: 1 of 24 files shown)


nvukobratTT (approved):
One note to think about is how to structure these post-processing steps. I.e., having a single entry point for all models.

The tricky part here is to outline cases like:
- Simple results print
- Post-processing and print
- Post-processing and saving the image
- Decoding output tokens
- Etc.

Either way, let's discuss this topic as well @kmabeeTT @mrakitaTT before we push for full restructuring. For now, I'm approving this PR as we'll have to restructure those either way with the variants introduction.

Issue #29, opened 2025-06-25
Add missing dtype_override to squeezebert load_inputs() 


kmabeeTT:
Quick fix.. Helpful to have dtype_override here like all other models to make automated model discovery and test generation work (something new being explored that requires all models have same interface).  Missing here, plz add: https://github.com/tenstorrent/tt-forge-models/blob/main/squeezebert%2Fpytorch%2Floader.py#L44

PR #31, opened 2025-06-25
Add missing dtype_override to load_inputs() of nlp models

kamalrajkannan78:
### Ticket
https://github.com/tenstorrent/tt-forge-models/issues/29

### Summary 
This PR fixes https://github.com/tenstorrent/tt-forge-models/issues/29 by adding missing dtype_override to load_inputs() of nlp models added in [this PR](https://github.com/tenstorrent/tt-forge-models/pull/19)

### DIFF
--- a/dpr/pytorch/loader.py
+++ b/dpr/pytorch/loader.py
@@ -41,7 +41,7 @@ class ModelLoader(ForgeModel):
         return model
 
     @classmethod
-    def load_inputs(cls):
+    def load_inputs(cls, dtype_override=None):
         """Generate sample inputs for DPR model."""
 
         # Ensure tokenizer is initialized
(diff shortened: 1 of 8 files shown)


kmabeeTT (dismissed):
Thanks!   Edit: Though hmm, maybe if dtype_override is not actually used like in this case, it makes sense not to add.  I realized we can workaround it in test using:

```
        def _load_inputs(self):
            # Check if load_inputs method supports dtype_override parameter
            sig = inspect.signature(self.loader.load_inputs)
            if "dtype_override" in sig.parameters:
                return self.loader.load_inputs(dtype_override=torch.bfloat16)
            else:
                return self.loader.load_inputs()
```

kmabeeTT:
I wouldn't merge this in it's current state since dtype_override is not used - can the dtype_override you added be used in the load_inputs() functions for these models?

kmabeeTT:
Can close, not strictly required to add these args.

Issue #27, opened 2025-06-24
Use variant enums instead of variant_name strings in models


kmabeeTT:
Opening ticket for further discussion..

@mrakitaTT originally proposed this here: https://github.com/tenstorrent/tt-forge-models/pull/7#discussion_r2158860332 - I held off on including in that PR since didn't want to introduce risk, and also since discussion was ongoing.   Recapping suggestion:

> Maybe we should create an enum like this in each loader:
> ```
> class Variants(StrEnum):
>     BASE = "albert-base-v2"
>     LARGE = "albert-large-v2"
>     XLARGE = "albert-xlarge-v2"
>     XXLARGE = "albert-xxlarge-v2"
> ```
> so we don't have to reference variants by strings from frontends, and you can use this enum for _VARIANTS and DEFAULT_VARIANT below..

@nvukobratTT said:

> What is the value of introducing enums for every model variant?
> In the majority of cases, this will be a 1-1 mapping. I.e., we'll have # enums == # variants. This looks like additional boilerplate code that we need to maintain.

Marko replied:

> Because variant names can be used from frontends to call some of the APIs here, so it would be better to have them extracted somewhere instead of using plain strings from frontends, for example: loader.get_model_info(Variants.BASE) instead of loader.get_model_info("albert-base-v2"). That way if someone changes variant name in tt-forge-models repo they won't break the frontends.

Since then I talked to @jameszianxuTT and he liked enum idea for variants too instead of strings for usage in tt-torch for setting per-variant FE specific params (pcc checking, pcc threshold, etc).

Yesterday I dropped [comment](https://github.com/tenstorrent/tt-forge-models/pull/7#discussion_r2162853492) some WIP changes for this to tt-forge-models branch [kmabee/model_variants_followup](https://github.com/tenstorrent/tt-forge-models/tree/kmabee/model_variants_followup) and some in tt-torch here: https://github.com/tenstorrent/tt-torch/tree/kmabee/model_variants_followup

jameszianxuTT:
For more context: 

I agree with enum idea, it seems like a less fragile solution than strings, as we will be attaching variant-specific + FE-specific info to tests in tt-torch. 

For example, we may want to set a specific PCC threshold for some model variants. This is FE specific data, so we shouldn't put it into the common tt-forge-models ModelInfo class.

[...]

PR #43, opened 2025-07-16
Use variant enums instead of variant_name strings in existing models and apply BASE variant to few models

kmabeeTT:
### Ticket
#27 

### Problem description

- We want to switch to use variant enums for type safety in models 
- We want to roll out variant support in more models

### What's changed

Here are some initial changes for review/discussion (@mrakitaTT @nvukobratTT plz take a quick peak) before rolling out to wider set of models

- Replace variant str with Variant(StrEnum) class in bunch of places, update doc strings
- Update models that have variants today (Bert, Albert) w/ changes
- get_model_info() taking str variant is still allowed for now to preserve back-compat until all models updated w/ this change
- Can run tests modified here (or those not modified) without any changes in tt-torch, backwards compatible
- Update yolov3, yolov4 to have variant support, single BASE variant for now, as illustrative example before rolling out similar change to all models.

I propose to submit this batch first then someone can pick up wholesale update of all models to add variant support with appropriate details.

### Checklist
- [x] Run tests locally in tt-torch (albert, bert, yolov3, yolov3 and unmodified model oft)

### DIFF
--- a/albert/masked_lm/pytorch/__init__.py
+++ b/albert/masked_lm/pytorch/__init__.py
@@ -2,4 +2,4 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from .loader import ModelLoader
+from .loader import ModelLoader, Variant
--- a/albert/masked_lm/pytorch/loader.py
+++ b/albert/masked_lm/pytorch/loader.py
@@ -16,7 +16,17 @@ from ....config import (
     ModelTask,
     ModelSource,
     Framework,
+    StrEnum,
 )
+
+
+class Variant(StrEnum):
+    """Available ALBERT model variants."""
+
+    BASE = "albert-base-v2"
+    LARGE = "albert-large-v2"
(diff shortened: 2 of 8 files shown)


kmabeeTT:
Small change I was on the fence about yesterday, I'm going to rename `Variant` to `ModelVariant` (lines up better with other class names) and use it as the type hint/doc-string in loader.py instead of base class `StrEnum`.

```
from ....config import (
    LLMModelConfig,
    ModelInfo,
    ModelGroup,
    ModelTask,
    ModelSource,
    Framework,
    StrEnum,
)


class Variant(StrEnum): <== Will rename to ModelVariant
    """Available ALBERT model variants."""

    BASE = "albert-base-v2"
    LARGE = "albert-large-v2"
    XLARGE = "albert-xlarge-v2"
    XXLARGE = "albert-xxlarge-v2"


class ModelLoader(ForgeModel):
```

PR #48, opened 2025-07-22
update yolov10 load_input function to load dataset

ctr-pmuruganTT:
### Problem description
Currently, `yolo` models retrieve input image from an url. To make it more robust, huggingface `datasets` package can be used. If this change is accepted, the same will be changed for remaining yolo models

### What's changed
* `load_inputs()` is modified to use the datasets library to load an input image
* `cppe-5` dataset is used from datasets package
* removed import statements of cv2, numpy and get_file as they are not used now

### Checklist

Tested these changes in tt-xla repo and attached the logs below

[yolov10data.log](https://github.com/user-attachments/files/21366851/yolov10data.log)
[Shape of the image before and after the change](https://github.com/user-attachments/files/21371773/shape_yolo.log)

### DIFF
--- a/yolov10/pytorch/loader.py
+++ b/yolov10/pytorch/loader.py
@@ -5,9 +5,9 @@
 YOLOv10 model loader implementation
 """
 import torch
-import cv2
-import numpy as np
-from ...tools.utils import get_file
+from torchvision import transforms
+from datasets import load_dataset
+
 from ...config import (
     ModelInfo,
     ModelGroup,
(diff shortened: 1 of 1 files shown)


yolov10/pytorch/loader.py
```diff
@@ -93,15 +93,24 @@ def load_inputs(self, dtype_override=None):
...
+        transform = transforms.Compose(
+            [
+                transforms.Resize((480, 640)),
+                transforms.ToTensor(),
+            ]
+        )
 
-        img = cv2.resize(img, (640, 480))  # Resize to model input size
-        img = img / 255.0  # Normalize to [0,1]
-        img = np.transpose(img, (2, 0, 1))  # HWC to CHW format
```
mrakitaTT, line 102: We should probably add these two transforms too to the `transforms.Compose` list above if this is the data preprocessing format that yolov10 expects, can you please check?

yolov10/pytorch/loader.py
```diff
@@ -93,15 +93,24 @@ def load_inputs(self, dtype_override=None):
...
+            "cppe-5", split="test"
+        )  # cppe-5 is a dataset of 5 classes for Combined Personal Protective Equipment
+
+        # Get first image from dataset
+        image = dataset[0]["image"]
+
+        # Preprocess the image
+        transform = transforms.Compose(
+            [
+                transforms.Resize((480, 640)),
```
mrakitaTT, line 107: Did you purposely invert dimensions here (`(480, 640)` vs `(640, 480)`) or this is the difference between torchvision resize and opencv resize?

yolov10/pytorch/loader.py
ctr-pmuruganTT: `.ToTensor()` does the following things

-  Normalizing the values to [0,1]
-  Converts HWC to CHW format
-  Converts a PIL Image or NumPy array into a PyTorch tensor

So I have replaced those 2 lines with just `transforms.ToTensor()` statement

yolov10/pytorch/loader.py
ctr-pmuruganTT: Yes. I have purposefully inverted the dimensions because  `PIL/transforms uses (height, width)` and `cv2 uses (width, height)`

ctr-pmuruganTT:
I have also added a log in the description which tracks the shape of the image tensor before and after the change

PR #69, opened 2025-08-05
Update Autoencoder postprocess to use PIL

ctr-pmuruganTT:
### Ticket
As stated in this [PR](https://github.com/tenstorrent/tt-xla/pull/847) removing the usage of Matplotlib and used PIL instead.

### Problem description
Postprocessing function of `Autoencoder` model uses `Matplotlib.pyplot` to save the image which is a larger library in size. In order to avoid using such a larger package, `PIL` is used to save the image

### Checklist

Tested these changes with the test in tt-xla repository. PFA log for reference
[autoencoder.log](https://github.com/user-attachments/files/21598991/autoencoder.log)

### DIFF
--- a/autoencoder_linear/pytorch/loader.py
+++ b/autoencoder_linear/pytorch/loader.py
@@ -8,7 +8,7 @@ Autoencoder Linear model loader implementation
 import os
 import numpy as np
 import torch
-import matplotlib.pyplot as plt
+from PIL import Image
 import torchvision.transforms as transforms
 from datasets import load_dataset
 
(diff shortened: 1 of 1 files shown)


mrakitaTT (approved):

mrakitaTT:
I guess you've already tested if this works correctly, but just to point out that there is a demo script in tt-forge repo which uses this: https://github.com/tenstorrent/tt-forge/blob/main/demos/tt-forge-fe/cnn/autoencoder_linear.py

PR #74, opened 2025-08-06
Remove unnecessary llama dtype_override

gengelageTT:
### Problem description
Setting overriding the type of all the input keys like this:

            if dtype_override is not None:
                        for key in inputs:
                            inputs[key] = inputs[key].to(dtype_override)
                            
Causes an error in tt-torch tests since not all input keys can be set to `torch.bfloat16`. Since this is an error in the forward pass and not in the tt-torch model tester, this must be changed in order for the model to pass with `dtype_override=torch.bfloat16`.

            E       RuntimeError: Expected tensor for argument #1 'indices' to have one of the following scalar types: Long, Int; but                         
            got CPUBFloat16Type instead (while checking arguments for embedding)

### What's changed
Removed these lines to allow the llama models to pass in tt-torch without error.

### Checklist

### DIFF
--- a/llama/causal_lm/pytorch/loader.py
+++ b/llama/causal_lm/pytorch/loader.py
@@ -240,9 +240,4 @@ class ModelLoader(ForgeModel):
         for key in inputs:
             inputs[key] = inputs[key].repeat_interleave(batch_size, dim=0)
 
-        # Only convert dtype if explicitly requested
-        if dtype_override is not None:
-            for key in inputs:
-                inputs[key] = inputs[key].to(dtype_override)
-
         return inputs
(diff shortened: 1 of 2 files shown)


llama/causal_lm/pytorch/loader.py
```diff
@@ -240,9 +240,4 @@ def load_inputs(self, dtype_override=None, batch_size=1):
         for key in inputs:
             inputs[key] = inputs[key].repeat_interleave(batch_size, dim=0)
 
-        # Only convert dtype if explicitly requested
-        if dtype_override is not None:
-            for key in inputs:
-                inputs[key] = inputs[key].to(dtype_override)
```
AleksKnezevic, line 246: Sorry for the late review, but would something like 
```
if torch.is_floating_point_dtype(dtype) and inputs[key].is_floating_point() or not torch.is_floating_point_dtype(dtype) and not inputs[key].is_floating_point():
  inputs[key] = inputs[key].to(dtype)
```

PR #76, opened 2025-08-07
detr, oft and vovnet ONNX models

ssaliceTT:
### Ticket
https://github.com/tenstorrent/tt-torch/issues/920

### Problem description
Model tests are being changed to be compatible with tt-forge-models. Models need to be added to tt-forge-models for this.

### What's changed
Added directories and files for dear, oft and VoVNet for ONNX.

### Checklist
- [ ] Check if we should have separate src files for onnx from pytorch in oft
- [ ] Decide if we should use timm for loader of VoVNet models or pytorchcv as done for pytorch
- [ ] Test VoVNet and make sure it works locally
- [ ] Run Nightly for three models

Nightly:

### DIFF
--- a/detr/onnx/__init__.py
+++ b/detr/onnx/__init__.py
@@ -4,4 +4,4 @@
 """
 DETR ONNX model implementation for Tenstorrent projects.
 """
-from .loader import ModelLoader
+from .loader import ModelLoader, ModelVariant
--- a/oft/onnx/__init__.py
+++ b/oft/onnx/__init__.py
@@ -4,4 +4,4 @@
 """
 OFT (Orthographic Feature Transform) ONNX model implementation for Tenstorrent projects.
 """
-from .loader import ModelLoader
+from .loader import ModelLoader, ModelVariant


kmabeeTT:
Let's think about this more.  Some initial thoughts:

- we haven't run OFT through onnx before, test_oft_onnx.py is untested in tt-torch, not sure what we do about this yet, we definitely don't want to duplicate src, and would rather run from pure onnx file hosted on s3 bucket.

- vovnet onnx, would rather keep the same variant (from timm) that we target from tt-torch I think not just make a copy of the existing pytorch version and make it export/load from onnx

I think we should hold off from porting any onnx models for now, since we don't have good plan here.  Originally the idea was to just load onnx files from s3/lfs bucket and run (like centernet model) but these ones will run pytorch, export to onnx, load from onnx, which is good for bringup but kind of defeats the purpose of the onnx tests.

nvukobratTT:
Moving to draft as we're still pushing major effort through PyTorch.

Issue #84, opened 2025-08-12
Add requirements.txt per model and move imports into load_model() to avoid at test-discovery time


kmabeeTT:
**Intro:**

We've wanted for a while to move python requirements out of top level project (like tt-torch, tt-forge-fe, tt-xla) `$ROOT/requirements.txt` and have them be per-model specific `requirements.txt` in this repo `tt-forge-models` , as in:

```
third_party/tt_forge_models/yolov6/pytorch/requirements.txt
third_party/tt_forge_models/yolox/pytorch/requirements.txt
etc
```

Now that `test_models.py` exists in tt-torch (and very soon tt-xla, works but not yet released) to automatically discover and run all models from `tt-forge-models`, it's worth revisiting this.

**Request:**

This request is two part:

1. Add appropriate per-model `requirements.txt` 
2. Move heavy imports from top of models `loader.py` global scope to be inside `load_model()` or near usage.

Number 1 is to be used by running project (`test_models.py` has logic not yet released that will install model specific requirements if found before running test, then restore after running test).

Number 2 is so that `test_models.py` can import loader.py files during model discovery but not have to install dependencies then, only when a particular model is run.

kmabeeTT:
Here is example of what would happen if model relies on a dependency installed via model-specific `requirements.txt` during runtime, and if we don't move imports into `load_model()` for example with yolov6:


```
pytest -svv tests/runner/test_models.py::test_all_models[yolov6/pytorch-yolov6n-full-eval] 

============================= test session starts ==============================
platform linux -- Python 3.10.12, pytest-8.4.1, pluggy-1.6.0 -- /localdev/kmabee/tt-torch/env/venv/bin/python3.10
cachedir: .pytest_cache
rootdir: /localdev/kmabee/tt-torch
configfile: pytest.ini
plugins: forked-1.6.0, cov-6.2.1, split-0.10.0, xdist-3.8.0
collecting ... ERROR: not found: /localdev/kmabee/tt-torch/tests/runner/test_models.py::test_all_models[yolov6/pytorch-yolov6n-full-eval]
(no match in any of [<Module test_models.py>])

Cannot import path: /localdev/kmabee/tt-torch/third_party/tt_forge_models/yolov6/pytorch/loader.py: No module named 'yolov6.layers'
collected 0 items

============================ no tests ran in 3.30s =============================
```

Solution to move the imports so they aren't needed at discovery:

```
[...]

PR #85, opened 2025-08-13
Update handful of models to have requirements.txt and defer imports to runtime

kmabeeTT:
### Ticket
#84 

### Problem description
 - We want to specify models dependencies in model specific requirements.txt so can be eliminated from project requirements.txt
 - And also defer importing to runtime to not be needed for model discovery via test_models.py

### What's changed
 - Here are some example changes covering handful of obvious cases 
 - yolov6, yolox, mlp_mixer, gliner, seamless_m4t, bi_rnn_crf are updated
 - Add requirements.nodeps.txt for yolox, to be installed with --no-deps

### Checklist
- [x] Modified models can run with test_models.py in tt-torch now without project installing deps in global requirements.txt when used with upcoming change to automatically install/uninstall deps for tt-forge-models models


yolox/pytorch/loader.py
```diff
@@ -5,19 +5,12 @@
 """
 YOLOX model loader implementation
 """
-import subprocess
-
-subprocess.run(
-    ["pip", "install", "yolox==0.3.0", "--no-deps"]
```
kmabeeTT, line 11: This was only model that did a pip-intall within the test source, that will be replaced with requirements.nodeps.txt here - since tt-forge-fe runs this model from hand written test runner, my hope is they can move this pip-install command there temporarily, or to their requirements.txt.  in tt-torch and tt-xla via test_models.py it will be automatically installed via the new requirements.txt and requirements.nodeps.txt added here.

nvukobratTT:
cc @saiarthiraguram

PR #100, opened 2025-08-20
Add JAX models for GPT2 and GPT-J

ctr-pmuruganTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/1049

### Problem description
Add JAX models for GPT-J and GPT-2 

### Checklist

### DIFF
--- a/gpt_j/causal_lm/jax/loader.py
+++ b/gpt_j/causal_lm/jax/loader.py
@@ -1,0 +1,146 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""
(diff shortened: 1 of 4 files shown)


gpt_j/causal_lm/jax/loader.py
```diff
@@ -0,0 +1,146 @@
...
+    ModelSource,
+    Framework,
+    StrEnum,
+)
+
+
+class ModelVariant(StrEnum):
+    """Available GPT-J model variants."""
+
+    GPT_J_6B = "gpt_j_6b"
```
mrakitaTT, line 26: We agreed to not use model name in the variant names so it doesn't get repeated when combining into full model name

mrakitaTT (approved):

PR #106, opened 2025-08-21
Add JAX models for MarianMT and LongT5

ctr-pmuruganTT:
### Ticket
 https://github.com/tenstorrent/tt-xla/issues/1071

### Problem description
Add JAX models for MarianMT and LongT5

### Checklist

PFA logs for reference :

[lt5_large_local_cg_main.log](https://github.com/user-attachments/files/21918553/lt5_large_local_cg_main.log)
[lt5_xl_tglobal_cg_main.log](https://github.com/user-attachments/files/21918555/lt5_xl_tglobal_cg_main.log)
[marianMT.log](https://github.com/user-attachments/files/21918557/marianMT.log)
[t5_base_cg_main.log](https://github.com/user-attachments/files/21918558/t5_base_cg_main.log)

### DIFF
--- a/longt5/text_classification/jax/loader.py
+++ b/longt5/text_classification/jax/loader.py
@@ -1,0 +1,163 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+LongT5 model loader implementation for Conditional Generation.
+"""
+
+from typing import Optional
+
+from ....base import ForgeModel
+from ....config import (
+    LLMModelConfig,
+    ModelInfo,
+    ModelGroup,
+    ModelTask,
+    ModelSource,
+    Framework,
+    StrEnum,
+)
+from transformers.models.longt5.modeling_flax_longt5 import shift_tokens_right
+
+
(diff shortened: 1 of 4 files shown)


longt5/text_classification/jax/loader.py
```diff
@@ -0,0 +1,162 @@
...
+        if self._tokenizer is None:
+            self._load_tokenizer(dtype_override)
+
+        # Load sample inputs
+        inputs_dict = self._tokenizer(
+            self.sample_text,
+            return_tensors="jax",
+        )
+
+        model = self.load_model()
```
mrakitaTT, line 147: This will cause the model to be loaded twice in our tests, once when we call `load_model()` and then again when we call `load_inputs()`. Since you only need the model config parameters here, you can load only the [LongT5Config](https://huggingface.co/docs/transformers/en/model_doc/longt5#transformers.LongT5Config) from `self._model_name` instead of loading the whole model. Other option would be to cache the model similar to `self._tokenizer`.

ctr-pmuruganTT: Updated the statement to use config params from `LongT5Config`

mrakitaTT (approved):

PR #114, opened 2025-08-25
update jax mlpmixer and mnist

ctr-lelanchelianTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/1114

### Problem description
- Add mlp_mixer and mnist Jax models in tt-forge-models
- MNIST is a custom model — do we need to use images from the Hugging Face dataset for that as well?

### Checklist

I have attached the logs for your reference:
[test_mlpmixer_without_xfail.log](https://github.com/user-attachments/files/21972017/test_mlpmixer_without_xfail.log)

### DIFF
--- a/dinov2/image_classification/jax/__init__.py
+++ b/dinov2/image_classification/jax/__init__.py
@@ -1,0 +1,4 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+from .loader import ModelLoader, ModelVariant
--- a/dinov2/image_classification/jax/loader.py
+++ b/dinov2/image_classification/jax/loader.py
@@ -1,0 +1,138 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""
+DINOv2 model loader implementation for image classification.
+"""
+
+from typing import Optional
+from ....base import ForgeModel
+from ....config import (
+    ModelConfig,
+    ModelInfo,
+    ModelGroup,
+    ModelTask,
+    ModelSource,
+    Framework,
+    StrEnum,
+)
+
+
+class ModelVariant(StrEnum):
+    """Available DINOv2 model variants."""
+
+    BASE = "base"
+    GIANT = "giant"
+    LARGE = "large"
+
+
+class ModelLoader(ForgeModel):
+    """DINOv2 model loader implementation for image classification."""
+
+    _VARIANTS = {
+        ModelVariant.BASE: ModelConfig(
+            pretrained_model_name="facebook/dinov2-base",
+        ),
+        ModelVariant.GIANT: ModelConfig(
+            pretrained_model_name="facebook/dinov2-giant",
+        ),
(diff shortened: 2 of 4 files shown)


mlp_mixer/image_classification/jax/loader.py
```diff
@@ -0,0 +1,159 @@
...
+    Framework,
+    StrEnum,
+)
+from .src.model_implementation import MlpMixer
+
+
+class ModelVariant(StrEnum):
+    """Available MLP Mixer model variants."""
+
+    MIXER_B16 = "mixer_b16"
```
mrakitaTT, line 32: Change to `BASE_16 = "base_16"`

mlp_mixer/image_classification/jax/loader.py
```diff
@@ -0,0 +1,159 @@
...
+
+        Args:
+            dtype_override: Optional dtype to override the default dtype.
+
+        Returns:
+            PyTree: Model parameters loaded from pretrained weights
+        """
+        # Download and load pretrained weights
+        link = "https://storage.googleapis.com/mixer_models/imagenet21k/Mixer-B_16.npz"
+        with fsspec.open("filecache::" + link, cache_storage="/tmp/files/") as f:
```
mrakitaTT, line 156: @vvukomanTT can this cause issues in CIv2? This is how this test was already run in tt-xla so I guess not, but I am not sure if this is a good practice to cache like this, I guess we should use S3 utilities here.

mnist/image_classification/jax/__init__.py
```diff
@@ -0,0 +1,8 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from .loader import ModelLoader, ModelArchitecture
+from .mlp.model_implementation import MNISTMLPModel
+from .cnn_nodropout.model_implementation import MNISTCNNNoDropoutModel
+from .cnn_dropout.model_implementation import MNISTCNNDropoutModel
```
mrakitaTT, line 8: No need to expose these

mnist/image_classification/jax/cnn_nodropout/model_implementation.py
```diff
@@ -0,0 +1,36 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from flax import linen as nn
+
+
+class MNISTCNNNoDropoutModel(nn.Module):
+    """MNIST CNN model implementation without dropout."""
```
mrakitaTT, line 9: Let's rename this to `MNISTCNNBatchNormModel`, and update the comment here and also the folder name and variant name.

mlp_mixer/image_classification/jax/loader.py
vvukomanTT: @mrakitaTT 
It won't cause any issues in CIv2 or CIv1, but it won't actually utilize caching at all.
It will download the weights to `/tmp/files` and we never read from that directory AFAIK (i.e. no cache reads).

If we want to utilize caching properly for this weights file, we can add it to the S3 bucket and then both CIv1 and CIv2 can use the file.

Issue #112, opened 2025-08-25
Misc models cleanup issues


kmabeeTT:
Creating a tracking issue for various small things spotted while reviewing models coverage in tt-forge-models

**unet-brain**

Same model twice, first version added by @gengelageTT then later version was added 3 weeks later by @meenakshiramanathan1 

```
unet/pytorch/loader.py:            hub_repo="mateuszbuda/brain-segmentation-pytorch",
unet/pytorch/loader.py:                "https://github.com/mateuszbuda/brain-segmentation-pytorch/raw/master/assets/TCGA_CS_4944.png",
unet/torch_hub/pytorch/loader.py:            pretrained_model_name="mateuszbuda/brain-segmentation-pytorch",
unet/torch_hub/pytorch/loader.py:            "https://github.com/mateuszbuda/brain-segmentation-pytorch/raw/master/assets/TCGA_CS_4944.png"

git blame unet/torch_hub/pytorch/loader.py | grep "mateuszbuda/brain-segmentation-pytorch"
3e118802 (Grace Engelage 2025-07-30 14:18:26 -0400  38)             pretrained_model_name="mateuszbuda/brain-segmentation-pytorch",
3e118802 (Grace Engelage 2025-07-30 14:18:26 -0400 110)             "https://github.com/mateuszbuda/brain-segmentation-pytorch/raw/master/assets/TCGA_CS_4944.png"

git blame unet/pytorch/loader.py | grep "mateuszbuda/brain-segmentation-pytorch"
80a53aec (Meenakshi Ramanathan 2025-08-21 21:20:55 +0530  66)             hub_repo="mateuszbuda/brain-segmentation-pytorch",
80a53aec (Meenakshi Ramanathan 2025-08-21 21:20:55 +0530 147)                 "https://github.com/mateuszbuda/brain-segmentation-pytorch/raw/master/assets/TCGA_CS_4944.png",
```

mrakitaTT:
We should also go through all models and add variants enum, and check/update the model names in `ModelInfo`. For example pixtral has this as a model name `"mistral-community/pixtral-12b"`.

PR #131, opened 2025-09-02
Remove duplicate unet loader

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-forge-models/issues/112

### Problem description
There are two nearly identical model loaders for unet (`mateuszbuda/brain-segmentation-pytorch`). 

### What's changed
Removed the version in the `/torch_hub` folder. Only notable differences between the two loaders was the model type and the method used to increase the batch size. Both methods to increase the batch size work and have been tested in tt-torch locally.

### Checklist


kmabeeTT (approved):
Great - thanks for comparing and removing!

PR #132, opened 2025-09-03
Add bge m3 model loader

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-torch/issues/920

### Problem description
Add model loader for bge_m3 from tt-torch.

### What's changed
- Added bge_m3 loaders for `model.model` and `model.encode_single_device`
- Added FlagEmbedding to requirements.txt for each model loader

### Checklist


bge_m3/encode/pytorch/loader.py
```diff
@@ -0,0 +1,104 @@
...
+    ModelSource,
+    Framework,
+    StrEnum,
+)
+
+
+class ModelVariant(StrEnum):
+    """Available BGE M3 variants."""
+
+    BGE_M3_ENCODE = "bge_m3_encode"
```
kmabeeTT, line 26: Don't love that model name and variant name are the same, the resulting model_info.name will end up containing both as "bg3_m3_encode...bge_m3_encode"  which Marko preferred we try and avoid blatant duplications -  Can you find a different way to avoid duplication.  Perhaps:

model: "bge_m3" variant: "encode"
although the pretrained_model_name is BAAI/bge-m3 and doesn't contain encode, so not the greatest either.

perhaps:

model: "bge_m3_encode" variant: BASE (base)

bge_m3/pytorch/requirements.txt
```diff
@@ -0,0 +1 @@
+FlagEmbedding
```
kmabeeTT, line 1: Nice! Thanks! This will get automatically installed and removed around running test via test_models.py (try it out and see in tt-torch)

AleksKnezevic (approved):

PR #155, opened 2025-09-19
UNIAD Pytorch Model Bringup

sonalibaskaran2499:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/1151

### Problem description

- Added the PyTorch implementation of the UNIAD model.
- The original implementation relied on external libraries such as mmcv, mmdet, and mmdet3d.
- Removed these dependencies by refactoring key functions, including config, DictAction, build_model, load_checkpoint, and set_random_seed.
- Replaced the use of build_dataset by directly passing randomly generated input tensors based on the shape and distribution of the NuScenes dataset.
- Substituted mmcv.load_checkpoint with model.load_state_dict() for loading model weights.

### What's changed
ModelLoader class for UNIAD has been added to the respective directories.

### Checklist

### DIFF
--- a/uniad/pytorch/__init__.py
+++ b/uniad/pytorch/__init__.py
@@ -1,0 +1,7 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""
+UNIAD PYTORCH model implementation for Tenstorrent projects.
+"""
+from .loader import ModelLoader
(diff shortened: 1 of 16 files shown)


kmabeeTT (approved):
Thanks for bringing up model - looks good to me, @AleksKnezevic @nvukobratTT @mrakitaTT any concerns?

kmabeeTT:
Ah, I didn't notice this, but this brings new dependency on package "casadi"

```
uniad/pytorch/src/planning_utils.py:from casadi import DM, Opti, OptiSol, sumsqr, vertcat, exp
````

which leads to:

```
Cannot import path: /localdev/kmabee/tt-xla/third_party/tt_forge_models/uniad/pytorch/loader.py: No module named 'casadi'
```

Even if was put in requirements.txt in uniad folder it wouldn't quite solve this problem since imports happen at top of file which affects model discovery, so don't have a choice here but to install casadi package in tt-xla environment for now...

Issue #171, opened 2025-09-26
Add support for TinyLlama


pmarkovicTT:
ML training team needs support for TinyLlama model as part of the generality push.

Source: TinyLlama/TinyLlama-1.1B-Chat-v1.0

PR #183, opened 2025-10-06
Add tinyllama support

pmarkovicTT:
### Ticket
#171 

### Problem description
ML training team needs TinyLlama model support for generality push.

### What's changed
Add new (TinyLlama) variant to the existing Llama loader

### Checklist


llama/causal_lm/pytorch/loader.py
```diff
@@ -225,10 +233,6 @@ def load_inputs(self, dtype_override=None, batch_size=1):
         Returns:
             dict: Input tensors suitable for causal LM.
         """
-
-        # Get max_length from the variant config
-        max_length = self._variant_config.max_length
```
pmarkovicTT, line 230: nit: not used so I removed it

PR #198, opened 2025-10-14
Add `unpack_output_training` to every `ModelLoader` to enable general fwd/bwd testing 

ndrakulicTT:
Add common output-unpacking for ModelLoader training

- Add `training_utils.py` with a small registry to unpack model outputs into a single `torch.Tensor`.
- Add `ForgeModel.unpack_forward_output()` that delegates to the registry.
- Register handlers for common HF outputs:
  - *...WithPast*, *...WithCrossAttentions* → `logits` / `last_hidden_state`
  - `Seq2SeqLMOutput`, `SequenceClassifierOutput*`, `TokenClassifierOutput` → `logits`
  - `ImageClassifierOutput*` → `logits`
  - `UNet2DConditionOutput` → `sample`
  - (and others used in repo)
- Keeps a clear escape hatch: models with ambiguous outputs can override `unpack_forward_output`.

### DIFF
--- a/training_utils.py
+++ b/training_utils.py
@@ -9,7 +9,7 @@ model output objects (e.g., Hugging Face model outputs).
 
 It maintains a registry of handlers that define how to unpack each output type.
 Intended to be used if the model returns a class that can be unambiguously unpacked into a single tensor that represents the whole output.
-In case model returns an ambiguous output (e.g, list/tuple) ModelLoader needs to override `unpack_forward_output`. 
+In case model returns an ambiguous output (e.g, list/tuple) ModelLoader needs to override `unpack_forward_output`.
 """
 import torch
 from typing import Any, Callable, Dict


FINDINGS.md
```diff
@@ -0,0 +1,319 @@
+# Findings: Model Output Formats for Remaining unpack_output_training Implementations
```
mmilosevicTT: If we include documentation generated by AI, please make sure to note that. It will be helpful for reviewers to know how to approach looking at those files, as they tend not to be structured in the most human-readable way.

dla/pytorch/loader.py
```diff
@@ -5,7 +5,8 @@
 DLA model loader implementation
 """
 
-from typing import Optional
+from typing import Optional, Any
+import torch
```
mmilosevicTT: nit: ditto about imports

regnet/pytorch/loader.py
```diff
@@ -5,11 +5,12 @@
 RegNet model loader implementation
 """
 
-from typing import Optional
+from typing import Optional, Any
 from dataclasses import dataclass
 from PIL import Image
 from torchvision import models, transforms
 import torch
+from transformers.modeling_outputs import ImageClassifierOutputWithNoAttention
```
mmilosevicTT: nit: could this go down next to other transformers imports?

base.py
```diff
@@ -159,6 +159,19 @@ def load_inputs(self, **kwargs):
...
+    @abstractmethod
+    def unpack_output_training(self, fwd_output: Any):
+        """Prepare output for backward pass
+        This function is called after the forward pass and before the backward pass.
+
+        Args:
+            fwd_output: Output from the forward pass
+
+        Returns:
+            Any: Tensor that can be fed to the backward pass
```
mmilosevicTT: In all files you say you return `torch.Tensor`, why are we making a difference here? It might be good also to add type annotation of the method if we know it is going to be `torch.Tensor`.

albert/masked_lm/pytorch/loader.py
```diff
@@ -183,6 +184,20 @@ def load_inputs(self, dtype_override=None):
...
+    def unpack_output_training(self, fwd_output: Any):
+        """Prepare output for backward pass.
+
+        Args:
+            fwd_output: Output from the forward pass
+
+        Returns:
+            torch.Tensor: Logits tensor for training
+        """
+        assert isinstance(
```
mmilosevicTT: If you plan to move all methods to a single file, please abstract this assert as a method where you pass expected and actual.

training_utils.py
```diff
@@ -0,0 +1,63 @@
...
+_register_attr("BaseModelOutputWithPast", "last_hidden_state")
+_register_attr("BaseModelOutputWithPastAndCrossAttentions", "last_hidden_state")
+_register_attr("CausalLMOutputWithCrossAttentions", "logits")
+_register_attr("CausalLMOutputWithPast", "logits")
+_register_attr("CLIPOutput", "logits_per_text")
+_register_attr("DepthEstimatorOutput", "predicted_depth")
+_register_attr("DPRReaderOutput", "end_logits")
+_register_attr("ImageClassifierOutput", "logits")
+_register_attr("ImageClassifierOutputWithNoAttention", "logits")
+_register_attr("LlavaCausalLMOutputWithPast", "logits")
```
nvukobratTT: Can we utilize general outputs instead of model-specific? I.e., does `LlavaCausalLMOutputWithPast ` have a base class that is used across multiple other models? Similar question for other model-focused outputs.

training_utils.py
```diff
@@ -0,0 +1,63 @@
...
+        v = getattr(output, _ATTR_REGISTRY[cls_name], None)
+        if isinstance(v, torch.Tensor):
+            return v
+        raise ValueError(f"{cls_name}.{_ATTR_REGISTRY[cls_name]} is not a torch.Tensor")
+
+    if cls_name in _FUNC_REGISTRY:
+        v = _FUNC_REGISTRY[cls_name](output)
+        if isinstance(v, torch.Tensor):
+            return v
+        raise ValueError(f"Custom func for {cls_name} did not return a torch.Tensor")
```
mmilosevicTT: nit: Let's keep error messages as proper sentences.
```suggestion
        raise ValueError(f"Custom function for {cls_name} did not return a torch.Tensor.")
```

training_utils.py
```diff
@@ -0,0 +1,63 @@
+import torch
+from typing import Any, Callable, Dict
+
+_ATTR_REGISTRY: Dict[str, str] = {}
+_FUNC_REGISTRY: Dict[str, Callable[[Any], torch.Tensor]] = {}
+
+def _register_attr(cls_name: str, attr: str) -> None:
```
mmilosevicTT: Is there really a need for both attr and func? You can always do a lambda internally that calls the attribute, and abstract lambda creation with a separate private method.

training_utils.py
ndrakulicTT: Checked it out, most of this are child classes of `ModelOutput` that dont have any default fields.
From this list only one that can be removed is `SegFormerImageClassifierOutput` because it is child class of `ImageClassifierOutput` and has same outputs

training_utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+    cls_name = output.__class__.__name__
+
+    if cls_name in _ATTR_REGISTRY:
+        v = getattr(output, _ATTR_REGISTRY[cls_name], None)
+        if isinstance(v, torch.Tensor):
+            return v
+        raise ValueError(f"{cls_name}.{_ATTR_REGISTRY[cls_name]} is not a torch.Tensor")
+
+    if cls_name in _FUNC_REGISTRY:
```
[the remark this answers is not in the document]
ndrakulicTT: Good point, will add check so that you cant insert same key twice either in func or args

training_utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+def _register_attr(cls_name: str, attr: str) -> None:
+    _ATTR_REGISTRY[cls_name] = attr
+
+def _register_func(cls_name: str, fn: Callable[[Any], torch.Tensor]) -> None:
+    _FUNC_REGISTRY[cls_name] = fn
+
+def unpack_forward_output(output: Any) -> torch.Tensor:
+    if isinstance(output, (tuple, list, dict)):
+        raise ValueError(
```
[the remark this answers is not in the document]
ndrakulicTT: There are models that not return classes. They will need to overwrite `ForgeModel::unpack_forward_output`

PR #210, opened 2025-10-23
Add requirements for Qwen2.5 Vl AWQ variants

kamalrajkannan78:
### Ticket
Link to Github Issue

### Problem description
Provide context for the problem.

### What's changed
Describe the approach used to solve the problem.
Summarize the changes made and its impact.

### Checklist

### DIFF
--- a/qwen_2_5_vl/pytorch/requirements.txt
+++ b/qwen_2_5_vl/pytorch/requirements.txt
@@ -1,0 +1,3 @@
+triton==3.3.0
+intel_extension_for_pytorch==2.7.0
+autoawq[cpu]==0.2.9


kmabeeTT:
While you're in the neighborhood plz also add variant for Qwen2.5-VL-72B , it was customer requested model just like 3B and 7B but is missing currently.  In tt-xla it has no chance to fit on single chip device, so can eventually have same status as other tests that are "too big" in tt-xla test config yaml when it lands there too for now until multichip support added later.

```
        status: NOT_SUPPORTED_SKIP
        reason: "Too large for single chip"
        bringup_status: FAILED_RUNTIME
```

PR #240, opened 2025-11-04
Fix Runtime Error in Surya OCR Model

meenakshiramanathan1:
### Ticket
#243 

### Problem description

-  Surya OCR Model faces `RuntimeError: Can't change device type after XLA runtime is initialized`.

### What's changed
The above runtime error is fixed by [this](https://github.com/tenstorrent/tt-xla/pull/1970) PR, post which model fails with below error
```
E       TypeError: to() received an invalid combination of arguments - got (method), but expected one of:
E        * (torch.device device = None, torch.dtype dtype = None, bool non_blocking = False, bool copy = False, *, torch.memory_format memory_format = None)
E        * (torch.dtype dtype, bool non_blocking = False, bool copy = False, *, torch.memory_format memory_format = None)
E        * (Tensor tensor, bool non_blocking = False, bool copy = False, *, torch.memory_format memory_format = None)
```
Passing device=device into the predictors ensures they store a concrete device , so all internal .to(device) calls receive a valid argument instead of a method. 

### Logs
[surya_ocr_text1.log](https://github.com/user-attachments/files/23618444/surya_ocr_text1.log)

### DIFF
--- a/suryaocr/pytorch/loader.py
+++ b/suryaocr/pytorch/loader.py
@@ -80,6 +80,7 @@ class ModelLoader(ForgeModel):
         """Initialize ModelLoader with specified variant."""
         super().__init__(variant)
         self._transform = transforms.Compose([transforms.ToTensor()])
+        self.image_tensor = None
 
     def load_model(self, dtype_override=None) -> nn.Module:
         """Load Surya OCR wrapper model.
@@ -92,6 +93,8 @@ class ModelLoader(ForgeModel):
             raise ImportError(
                 "Surya package is not available. Please install `surya` to use SuryaOCR loader."
             )
+        if self.image_tensor is None:
+            self.load_inputs()
         if self._variant == ModelVariant.OCR_TEXT:
             model = SuryaOCRWrapper(image_tensor=self.image_tensor)
         elif self._variant == ModelVariant.OCR_DETECTION:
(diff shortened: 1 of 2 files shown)


kmabeeTT:
There is/was an ugly unsolved issue with surya-ocr exposed when adding `test/runner/test_models.py` in tt-xla back in August/September and I had to exclude the model from being discovered, which is recently in the file `tests/runner/utils/dynamic_loader.py` as follows:

```
        # TODO(kmabee) - Temporary workaround to exclude models with fatal issues.
        # Dictionary mapping model directory names to their exclusion reasons
        excluded_model_dirs = {
            "suryaocr": "Issue #1166"  # Surya OCR imports and initializes torch_xla runtime which causes issues
        }
```

The details in that linked issue are a bit light, but I recall it was causing a hang with --forked in any test due to surya-ocr package itself initializing torch_xla runtime.  Now I wonder if somehow you have bypassed this issue, or solved it?  Try removing above workaround and checking if any test_models.py tests work with pytest --forked.

meenakshiramanathan1:
@kmabeeTT Yes, the workaround was removed and the below runtime error
`RuntimeError: Can't change device type after XLA runtime is initialized` has been bypassed with the fixes on this PR

kmabeeTT:
Awesome - great stuff - thank you!

saiarthiraguram (approved):

PR #261, opened 2025-11-10
Moving torch mnist to tt_forge_models

ajakovljevicTT:
Moving torch_mnist to tt_forge_models fully, so it can be ran with `test_models.py` infra.

Deleted the existing mnist and replaces with the two types that were implemented in tt-xla. Chacked through tt-xla that they work.


mnist/image_classification/pytorch/loader.py
```diff
@@ -0,0 +1,193 @@
...
+            dtype_override: Optional torch.dtype to override the model's default dtype.
+                            If not provided, the model will use bfloat16.
+
+        Returns:
+            torch.Tensor: Input tensor that can be fed to the model.
+        """
+        # MNIST uses 28x28 grayscale images
+        # Batch size of 4, 1 channel (grayscale), 28x28 image
+        dtype = dtype_override if dtype_override is not None else torch.bfloat16
+        return torch.ones((4, 1, 28, 28), dtype=dtype)
```
kmabeeTT: Hmm not sure about using a tensor of all ones as inputs here in tt-forge-models.  I see the MNIST tests in tt-xla do this via https://github.com/tenstorrent/tt-xla/blob/main/tests/torch/single_chip/models/mnist/cnn/tester.py#L32 - but previous mnist in tt-forge-models was better and used datasets something I think we should keep (@mrakitaTT always pushed for increased datasets usage)

```
def load_input(batch_size=1):
    transform = transforms.Compose([transforms.ToTensor()])
    test_dataset = datasets.MNIST(
        root="./data", train=False, transform=transform, download=True
    )
    dataloader = DataLoader(test_dataset, batch_size=batch_size)
    test_input, _ = next(iter(dataloader))
    return test_input
```

PR #273, opened 2025-11-13
Add bias sharding for Qwen 2.5 models

hshahTT:
### Ticket
[tt-xla#2150](https://github.com/tenstorrent/tt-xla/issues/2150)

### Problem description and changes
The Qwen 2.5 models have attention biases so they should also be sharded to save memory.

### Checklist


kmabeeTT (approved):
Good catch, thanks!

PR #307, opened 2025-11-27
Skip BFloat16 conversion in ssdlite320_mobilenetv3 (nms_kernel lacks BFloat16 support)

kamalrajkannan78:
### Ticket

- Fixes https://github.com/tenstorrent/tt-xla/issues/2350

### Problem description

- ssdlite320_mobilenetv3 encounters `RuntimeError: "nms_kernel" not implemented for 'BFloat16'`

### Root cause

- This model uses [batched_nms](https://github.com/pytorch/vision/blob/59a3e1f9f78cfe44cb989877cc6f4ea77c8a75ca/torchvision/models/detection/ssd.py#L453), which is currently not supported in bfloat16.

### What's changed

- Skipped BFloat16 conversion for this model

### Checklist
- [x] Verified the changes through Local testing

### Logs

- [nov27_ssd_mob_before_fix.log](https://github.com/user-attachments/files/23796665/nov27_ssd_mob.log)
- [nov28_ssd_mob_after_fix.log](https://github.com/user-attachments/files/23820457/nov28_ssd_mob.log)

### DIFF
--- a/ssdlite320_mobilenetv3/pytorch/loader.py
+++ b/ssdlite320_mobilenetv3/pytorch/loader.py
@@ -78,7 +78,7 @@ class ModelLoader(ForgeModel):
 
         Args:
             dtype_override: Optional torch.dtype to override the model's default dtype.
-                           If not provided, the model will use its default dtype (typically float32).
+                           NOTE: This parameter is currently ignored (model always uses float32).
 
         Returns:
             torch.nn.Module: The SSDLite320 MobileNetV3 model instance.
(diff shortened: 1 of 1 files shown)


saiarthiraguram (approved):

vkovinicTT (changes requested):

kamalrajkannan78:
> In the fixed version of log we lave `vhlo` -> `shlo` -> `ttir` -> `ttnn` then again `vhlo` -> `shlo` -> `ttir`-> `ttnn`, why is this the case (it's repeated mulitple times). Shouldn't we just have each stage only once?

- The compilation stages (vhlo → stablehlo → ttir → ttnn) are NOT repeated for the same graph. Each sequence you see is for a different graph/subgraph. The stages appear multiple times because the model generates multiple separate subgraphs that each need to be compiled independently.
- This is normal behavior in TT-XLA. I've observed this in other models to

vkovinicTT (approved):

kmabeeTT (approved):
Approved - dtype_override argument to functions could have also been removed, but this is okay too, and more obvious.

PR #348, opened 2025-12-15
PCC drop fix for Gemma-1.1-7B 

sonalibaskaran2499:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/2344)

### Problem description
PCC drop was observed in the model, with the PCC value falling to 0.9790841341018677

### What's changed
Re-tested Gemma-1.1-7B-it  with padding = true and apply_chat_template:
pcc=1.0563201904296875

**Logs**
[gemma_1.1.7b-it.log](https://github.com/user-attachments/files/24190885/gemma_1.1.7b-it.log)
[gemma_2-2b-it.log](https://github.com/user-attachments/files/24197067/gemma_2-2b-it.log)
[gemma-1.1-2b-it.log](https://github.com/user-attachments/files/24197080/gemma-1.1-2b-it.log)
[gemma-2-9b-it.log](https://github.com/user-attachments/files/24197085/gemma-2-9b-it.log)

### Checklist

### DIFF
--- a/gemma/pytorch/loader.py
+++ b/gemma/pytorch/loader.py
@@ -42,11 +42,9 @@ class ModelLoader(ForgeModel):
     _VARIANTS = {
         ModelVariant.GEMMA_1_1_2B_IT: LLMModelConfig(
             pretrained_model_name=str(ModelVariant.GEMMA_1_1_2B_IT),
-            max_length=256,
         ),
         ModelVariant.GEMMA_1_1_7B_IT: LLMModelConfig(
             pretrained_model_name=str(ModelVariant.GEMMA_1_1_7B_IT),
-            max_length=256,
         ),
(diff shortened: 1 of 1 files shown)


kmabeeTT (approved):
Is the reason that we special case 7B_IT variant just because it's the one with low PCC? If the change makes sense for all variants I'm thinking actually we should apply it for all. Can you test all without and with this change locally or on CI for Blackhole and compare?

kmabeeTT:
Thanks for info @sonalibaskaran2499 and doing the comparisons - this looks great, in this case I think it makes sense to apply the chat_template just for it (instruction tunred) variants which from my understanding are the ones that benefit from chat_template. If you agree let's go with that.

PR #352, opened 2025-12-16
Disable warning regarding model info variants

ajakovljevicTT:
Since a lot of models do not handle variants properly, in tt-xla we get a lot of noise with the related warning. In agreement with @kmabeeTT, removing the warning.


kmabeeTT (approved):
Thx, it's needed now since we call the API for every model during collection, maybe will circle back and cleanup in future.

PR #353, opened 2025-12-16
initial load_inputs_decode() single decode-step changes for few models (llama, falcon, gemma, qwen)

kmabeeTT:
**Problem / Details:**

- We want to add decode tests for all of the LLMs we have running, but specifically for the multi chip LLMs. That entails adding a past cache input, and a generating a single token. The only difference is what inputs your loading, you need to pass in a cache tensor.

**Changes Made:**

- Add decode-only input path load_inputs_decode() to Llama, Falcon, Gemma, Qwen 2.5, and Qwen 3 loaders. These are just a few common models for now that are covered by single_device (n150/p150) and multichip tensor-parallel (n300-llmbox) in tt-xla CI.
- Decode path uses StaticCache (CPU), single-token input_ids, cache_position = max_cache_len - 1, use_cache=True; attention_mask omitted for single-batch decode.
 - Introduce shared helper get_simple_decode_token_id() (utils.py) to pick a deterministic non-special token and get_static_cache_decode_inputs() to construct Static Cache and input dict to reduce duplication across models' load_inputs_decode() function.
 - Keep existing load_inputs() unchanged (prefill-like); no behavioral changes to default tests. Sets groundwork for future phase selection (prefill vs decode) without changing tester infra.

**Testing:**

- Tested on tt-xla in models modified here via test_models.py changes to discover models with `load_inputs_decode()` function and create specialized tests for that path.


llama/causal_lm/pytorch/loader.py
```diff
@@ -284,6 +286,97 @@ def load_inputs(self, dtype_override=None, batch_size=1):
...
+        """
+        # Ensure tokenizer and config are initialized
+        if self.tokenizer is None:
+            self._load_tokenizer(dtype_override=dtype_override)
+        if getattr(self, "config", None) is None:
+            self.load_config()
+
+        # Create zero-initialized static cache for decode
+        cache_dtype = dtype_override if dtype_override is not None else torch.bfloat16
+        max_cache_len = self._variant_config.max_length
```
AleksKnezevic: I see we hardcode this to 128, probably useful to expose as an argument to the model/input loader, but not an issue for now.

llama/causal_lm/pytorch/loader.py
```diff
@@ -284,6 +286,97 @@ def load_inputs(self, dtype_override=None, batch_size=1):
...
+        static_cache = StaticCache(
+            config=self.config,
+            max_batch_size=batch_size,
+            max_cache_len=max_cache_len,
+            device="cpu",
+            dtype=cache_dtype,
+        )
+
+        # Single decode token per batch (use eos token id)
+        eos_id = self.tokenizer.eos_token_id
```
AleksKnezevic: why eos_token_id? We'll want to run a few decode loops, I think we do three warm ups and then the perf measurement ones, but that said, it will just loop eos_token, so probably fine.

llama/causal_lm/pytorch/loader.py
```diff
@@ -284,6 +286,97 @@ def load_inputs(self, dtype_override=None, batch_size=1):
...
+        input_ids = torch.full((batch_size, 1), fill_value=eos_id, dtype=torch.long)
+
+        # Minimal attention mask for the single token
+        attention_mask = torch.ones((batch_size, 1), dtype=torch.long)
```
AleksKnezevic: the attention mask should cover past cache size + 1, a little surprised this works actually... must get broadcast?

llama/causal_lm/pytorch/loader.py
kmabeeTT: Good point @jameszianxuTT, I added helper function in utils.py that other models can use which returns a non-special token, called via:

```
        token_id = get_simple_decode_token_id(self.tokenizer, self.config)
```

Take a look and let me know what you think?

falcon/pytorch/loader.py
```diff
@@ -258,3 +260,36 @@ def load_config(self):
         )
 
         return self.config
+
+    def load_inputs_decode(self, dtype_override=None, batch_size=1):
```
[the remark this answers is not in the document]
kmabeeTT: Yeah I too was concerned that it is the same code in every model, but figured at least it was minimal code, and the same patterns for load_inputs() are basically repeated in all model loader.py files by design, so that per-model tweaks can be made.

But I wasn't aware of fragility of StaticCache you raise... will see what this can look like in commonized function like you're describing - thx for suggestion!

AleksKnezevic (approved):

kmabeeTT:
Thx for reviews, going to merge this and put up tt-xla PR next that uses this.

PR #357, opened 2025-12-19
EasyDel implementation of Clip JAX model

ctr-pmuruganTT:
### Problem description
Modify loader of CLIP Jax model to use EasyDel Instead of Transformers library

### What's changed
* Modified the model source
* Removed `load_processor` and used AutoImageProcessor
* Added  `get_input_activations_partition_spec` and  `load_parameters_partition_spec`

### Checklist


clip/image_classification/jax/loader.py
```diff
@@ -115,14 +97,10 @@ def load_model(self, dtype_override=None):
             model: The loaded model instance
         """
 
-        from transformers import FlaxCLIPModel
+        from easydel import CLIPModel
```
vzeljkovicTT: Using ClipModel is good when we already have easydel weights, while in our case we are taking weights from HuggingFace and converting them into jax numpy arrays. Easydel gave us this AutoModel group of classes which is exactly doing that (params transfer to jax arrays and after that calling model implementation e.g. ClipModel). For this use case we should chose `AutoEasyDeLModelForZeroShotImageClassification` (`from easydel import AutoEasyDeLModelForZeroShotImageClassification`).

clip/image_classification/jax/loader.py
```diff
@@ -80,32 +84,10 @@ def _get_model_info(cls, variant: Optional[ModelVariant] = None) -> ModelInfo:
             variant=variant,
             group=ModelGroup.GENERALITY,
             task=ModelTask.MM_IMAGE_CAPT,
-            source=ModelSource.HUGGING_FACE,
+            source=ModelSource.EASYDEL,
             framework=Framework.JAX,
         )
 
-    def _load_processor(self, dtype_override=None):
```
vzeljkovicTT: Why did we deleted this _load_processor? It was better to use `AutoImageProcessor` instead of `CLIPProcessor`?

clip/image_classification/jax/loader.py
```diff
@@ -132,7 +110,7 @@ def load_model(self, dtype_override=None):
         from_pt = pretrained_model_name == "openai/clip-vit-large-patch14-336"
```
vzeljkovicTT: This flag from_pt is unnecessary in the easydel

nvukobratTT:
Staled PR, closing. Please reopen if still relevant.

PR #384, opened 2026-01-07
Add surya-ocr as a requirement for llama model

ctr-pmuruganTT:
### Problem description
`llama/causal_lm/pytorch-llama_3_1_8b-tensor_parallel-inference` variant is failing with an error in [PR ](https://github.com/tenstorrent/tt-xla/actions/runs/20770277201/job/59646563466?pr=2505) which removes surya-ocr from common venv

### What's changed
* Added surya-ocr pacakge as a requirement specific to this model

### Checklist

Tested this change in [CI](https://github.com/tenstorrent/tt-xla/actions/runs/20776467548/job/59664100649#logs) which is passing 
Without this change, [CI](https://github.com/tenstorrent/tt-xla/actions/runs/20771184501) is failing


kmabeeTT (changes requested):
Hmm this feels super weird to me, don't think it makes sense, and don't see why llama would need surya-ocr.  

I think it would be some other issue especially since error is:

```
>       assert (
            result.returncode == 0
        ), f"Failed to run multihost test for model variant: {model_variant}"
E       AssertionError: Failed to run multihost test for model variant: llama/causal_lm/pytorch-llama_3_1_8b-tensor_parallel-inference
E       assert -6 == 0
E        +  where -6 = CompletedProcess(args=['/__w/tt-xla/tt-xla/venv/bin/python', '-m', 'pytest', '-ssv', 'tests/runner/test_models.py::test_all_models_torch[llama/causal_lm/pytorch-llama_3_1_8b-tensor_parallel-inference]', '--disable-perf-measurement'], returncode=-6).returncode

```

Keep in mind when changing requirements in tt-xla, a new docker image is built (link from your PR run: https://github.com/tenstorrent/tt-xla/actions/runs/20770277201/job/59645008643?pr=2505) which may introduce other changes...somehow... maybe..

jameszianxuTT:
@kmabeeTT I had a look at this locally and something really weird is going on here... gonna dump a pip freeze w/ and w/o surya_ocr in the distributed test subprocs to see if something nonsensical is going on. 

It might be something related to a double import of some deps causing conflicts induced by how the distributed tests trigger test_models.py in a python subprocess... this doesn't explain why it's CI-only, though it might also be related to wheel building and package import resolution from a wheel vs source build. 

The failure in distributed is also not reproducible locally which makes things rather annoying, but we can also debug that further.

Anyways, the failure conditions are as follows:

```
llama 
only distributed multihost test
only in CI
only if surya-ocr is removed from requirements-dev python package dependencies
```

PR #427, opened 2026-01-22
Added GLM 4.5 and 4.7 models

AleksKnezevic:
Added loader for GLM models.

### DIFF
--- a/glm/causal_lm/pytorch/loader.py
+++ b/glm/causal_lm/pytorch/loader.py
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 """
-Llama model loader implementation for causal language modeling.
+Model loader implementation for causal language modeling.
 """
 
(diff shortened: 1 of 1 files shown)


glm/causal_lm/pytorch/loader.py
```diff
@@ -0,0 +1,296 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""
+Llama model loader implementation for causal language modeling.
```
kmabeeTT: Here and 9 other references to llama is probably copy/paste issue should be updated

AleksKnezevic: yeah, single layer runs on galaxy (first layer is dense, so no surprise), multiple layers fail to compile thorugh torch.compile due to shape dynamism.

PR #453, opened 2026-02-02
Add expanded (and explicit) support for prefill testing for Llama models

umalesTT:
This PR adds ```load_inputs_prefill``` function to ```loader.py``` for Llama models which will, analogous to ```load_inputs_decode``` prepare inputs for testing prefill in tt-xla infra, but in a more general way than only ```load_inputs``` - here different batch sizes and longer sequence lengths (with realistic input texts per sequence length) are explored.
 
Relevant issues for [sequence_length](https://github.com/tenstorrent/tt-xla/issues/3018) and [batch_size](https://github.com/tenstorrent/tt-xla/issues/3019).

For now, these changes are specific to Llama, but by relocating the added files we can easily support other models, which we will do in the near future.

Corresponding tt xla [branch](https://github.com/tenstorrent/tt-xla/tree/umales/llama_prefill_coverage).

### DIFF
--- a/llama/causal_lm/pytorch/loader.py
+++ b/llama/causal_lm/pytorch/loader.py
@@ -137,17 +137,21 @@ class ModelLoader(ForgeModel):
     # Sample text for causal LM
     sample_text = "Hey how are you doing today?"
 
-    def __init__(self, variant: Optional[ModelVariant] = None):
+    def __init__(
+        self, variant: Optional[ModelVariant] = None, num_layers: Optional[int] = None
+    ):
         """Initialize ModelLoader with specified variant.
 
         Args:
             variant: Optional ModelVariant specifying which variant to use.
                      If None, DEFAULT_VARIANT is used.
+            num_layers: Optional number of hidden layers to use. If None, uses the model's default.
         """
         super().__init__(variant)
         self.tokenizer = None
         self.seq_len = None
         self.config = None
+        self.num_layers = num_layers
 
     @classmethod
     def _get_model_info(cls, variant: Optional[ModelVariant] = None) -> ModelInfo:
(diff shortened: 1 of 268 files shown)


llama/causal_lm/pytorch/loader.py
```diff
@@ -322,6 +314,49 @@ def load_inputs_decode(self, dtype_override=None, batch_size=1):
...
+        # Get appropriate texts for this seq_len and batch_size
+        # Fall back to default seq_len if requested one is not available
+        if seq_len not in PREFILL_TEXTS:
+            available = list(PREFILL_TEXTS.keys())
+            # Use largest available that's <= seq_len, or smallest if seq_len is too small
+            candidates = [s for s in available if s <= seq_len]
+            fallback_seq_len = max(candidates) if candidates else min(available)
+            texts = get_prefill_texts_for_batch(fallback_seq_len, batch_size)
+        else:
+            texts = get_prefill_texts_for_batch(seq_len, batch_size)
```
pglusacTT: nit: You can do `seq_len = max(candidates) if candidates else min(available)` and not have an else branch.

kmabeeTT (approved):
Nice, looks awesome! A question and a small nit, dropping approval here in the meantime.

==

I checked out the tt-xla changes too on referenced branch a bit, and it looks cool. Future comment related to that:  I think the test-config yaml entries specifying batch-size and seq-len explicitly should find a way to be removed since that part is already determined by the test paramters, it seems duplicate/redundant):

```
  # Llama 3.2 1B Prefill tests with different seq_len and batch_size
  llama/causal_lm/pytorch-llama_3_2_1b-llm_prefill-seq128-batch1-single_device-inference:
    status: EXPECTED_PASSING
    seq_len: 128
    batch_size: 1
```

PR #463, opened 2026-02-09
Fix Torch Dynamo failures in SuryaOCR Model

meenakshiramanathan1:
### Ticket
Fixes[ #1193](https://github.com/tenstorrent/tt-xla/issues/1193)

### Problem description
Surya ocr variants face numpy related attribute errors.

### What's changed

-  Surya OCR model has custom test in tt-xla. Model faces lots of torch dynamo failures caused by the error arising from mixing NumPy arrays with Torch math inside a region that TorchDynamo is tracing. These can be fixed only by monkey-patching code from the surya-ocr package files. Hence all the patches has been moved to tt_forge_models instead of keeping it in tt-xla and the custom test from tt-xla can be removed later to enable runs through test_models_torch.

- Model was failing with AttributeError("'ndarray' object has no attribute 'mul'") . The original _image_processor used NumPy ops on a np.ndarray (e.g., image.astype(np.float64) * self.rescale_factor). Under TorchDynamo, this path got intercepted, and a Tensor-style .mul ended up being applied to an ndarray, triggering: AttributeError("'ndarray' object has no attribute 'mul'").My replacement converts the input to a Torch tensor up-front and performs same with Torch ops, only converting back to NumPy at the end. This keeps everything Torch-native while Dynamo traces, so no Tensor method is ever called on an ndarray.

- Following this, model failed with AssertionError: Mismatch in image embedding seq len: torch.Size([8192, 1280]) vs torch.Size([5584, 1280])I was not encountering this issue in cpu run, only xla run fails with above error.My patch disables the extra padding by setting extra_multipler = 1, so factor is minimal. The padded dimensions now match the resolution assumed by the positional encoding.

- Further model failed with cache part in dynamic ops with an assertion being hit.The patched init guards the call with if not torch._dynamo.is_compiling(), so no forbidden callable is executed during tracing.On fixing this, model is getting killed in xla. For now, this variant will be skipped and debugged later.

- Surya OCR detection varisnt fails after hitting this assertion assert isinstance(value.base, np.nditer). TorchDynamo is tracing through NumPy-based preprocessing in Surya’s detector. This call  routes into a  processor that operates on np.ndarray and iterator.Dynamo’s numpy interception hits an internal assertion involving np.nditer, causing the failure. Mixing NumPy-heavy preprocessing inside a Dynamo-traced path triggers numpy handling (np.nditer) assertion. Converting that preprocessing to torch-only ops fixed this issue.

- The latest runtime error is caused since XLA fuses or constant‑folds parts of the graph so an output from one branch can be  reused/broadcast against a tensor from another branch, triggering an illegal broadcast between incompatible shapes. Dependency has been added on the input so the output is no longer treated as a compile-time constant, so this way the output is made dependent on x without changing its value. 

- This prevents XLA from treating different outputs as the same constant and merging them together.After this change, model passes e2e.

- Verified the original implementation cpu results with model results after all changes and both were matching.

- After these changes, surya_ocr_text test gets killed which will be skipped for now while surya_ocr_detect test fails with out of memory error which will be added as a xfail in test config file.

### Logs
[surya_text_xla.log](https://github.com/user-attachments/files/25358821/surya_text_xla_thru_config_1.log)
[surya_detect_xla.log](https://github.com/user-attachments/files/25358824/surya_detect_xla_thru_config.log)

### DIFF
--- a/suryaocr/pytorch/src/utils.py
+++ b/suryaocr/pytorch/src/utils.py
@@ -4,13 +4,500 @@
 import json
 import os
 from collections import defaultdict
-from typing import List
 
 import torch
 import torch.nn as nn
 import torchvision.transforms as transforms
+import torchvision.transforms.functional as TF
 from loguru import logger
 from PIL import Image
+
+
+import math
+from typing import List, Optional, Union
+
+import cv2
+import numpy as np
+from surya.common.surya.processor import (
+    SuryaOCRProcessor,
+)
+from surya.detection.processor import (
+    SegformerImageProcessor,
+)
+from transformers import PretrainedConfig
+from transformers.image_utils import (
+    ChannelDimension,
+    ImageInput,
+    PILImageResampling,
+    infer_channel_dimension_format,
+)
+
+from surya.common.surya.processor import (
+    SuryaOCRProcessor,  # type: ignore[reportMissingImports]
+)
+from surya.detection import (
+    DetectionPredictor,
+)
+from surya.detection import (
+    heatmap as _surya_heatmap,  # type: ignore[reportMissingImports]
+)
+from surya.detection import heatmap as _surya_heatmap2
+from surya.detection.processor import (
+    SegformerImageProcessor,  # type: ignore[reportMissingImports]
+)
+from surya.foundation.cache.dynamic_ops import (
+    DynamicOpsCache,  # type: ignore[reportMissingImports]
+)
+from surya.foundation.cache.static_ops import (
+    StaticOpsCache,  # type: ignore[reportMissingImports]
+)
+from surya.settings import settings  # type: ignore[reportMissingImports]
+from surya.common.surya import SuryaModel  # type: ignore[reportMissingImports]
+
+
+def _prepare_image(self, img):
+    new_size = (self.processor.size["width"], self.processor.size["height"])
+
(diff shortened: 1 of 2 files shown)


saiarthiraguram (approved):

suryaocr/pytorch/src/utils.py
```diff
@@ -4,15 +4,471 @@
...
+    std = torch.as_tensor(
+        self.image_std, dtype=tensor_img.dtype, device=tensor_img.device
+    )
+
+    if tensor_img.ndim == 3 and tensor_img.shape[-1] == int(mean.numel()):
+        # HWC layout
+        tensor_img = (tensor_img - mean) / std
+    else:
+        # Fallback broadcast
+        tensor_img = (tensor_img - mean.view(1, 1, -1)) / std.view(1, 1, -1)
```
vkovinicTT: Shouldn't `torch` support automatic broadcast?

vkovinicTT (changes requested):

vkovinicTT:
In the `surya_text_xla.log` it has error where it says "_Backend compiler `inductor`_", shouldn't we use our custom backend and not cpu?

meenakshiramanathan1:
Closing as this model will be re-tested in latest main for current status

PR #472, opened 2026-02-10
Update model names to human-readable format for presentation (#447)

ctr-lelanchelianTT:
### Ticket
closes https://github.com/tenstorrent/tt-xla/issues/2970

### Problem description
Currently, model architecture and variant name are more in the form of encoding rather than the actual name. We should update them to represent human-readable names used for model presentation purposes

### What's changed
This PR updates them to use human-readable naming conventions. As part of this change, the model variant names have been corrected, and in some places where the ModelVariant class incorrectly used author names, those have been fixed as well. This aligns with the guidelines specified in the
[README](https://github.com/tenstorrent/tt-forge-models#:~:text=Always%20use%20a%20dedicated%20ModelVariant%20enum%20class%20with%20a%20DEFAULT_VARIANT%20class%20attribute) . I have tested the affected models individually, and the corresponding logs are attached for reference.

### DIFF
--- a/motif_image_6B_preview/requirements.txt
+++ b/motif_image_6B_preview/requirements.txt
@@ -1,1 +1,1 @@
-diffusers
+diffusers==0.36.0


nvukobratTT (approved):
Few small comments for changes I spot. Approving

PR #489, opened 2026-02-19
Fixes Runtime Issue in Bilstm-crf

### LINKED ISSUE #3319: BiLSTM RuntimeError: The tensor has a non-zero number of elements, but its data is not allocated yet.
### Describe the bug

- While testing the biLSTM model produces the Runtime error

```
E       RuntimeError: The tensor has a non-zero number of elements, but its data is not allocated yet.
E       If you're using torch.compile/export/fx, it is likely that we are erroneously tracing into a custom kernel. To fix this, please wrap the custom kernel into an opaque custom op. Please see the following for details: https://pytorch.org/tutorials/advanced/custom_ops_landing_page.html
E       If you're using Caffe2, Caffe2 uses a lazy allocation, so you will need to call mutable_data() or raw_mutable_data() to actually allocate memory.

```

### Steps to reproduce

`pytest tests/runner/test_models.py::test_all_models_torch[bi_lstm_crf/pytorch-default-single_device-full-inference -svv`

### Logs

[bilstm.txt](https://github.com/user-attachments/files/25356790/bilstm.txt)

saiarthiraguram:
### Ticket
Fixes: [Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/3319)

### Problem description
- While testing the biLSTM model produces the Runtime error

```
E       RuntimeError: The tensor has a non-zero number of elements, but its data is not allocated yet.
E       If you're using torch.compile/export/fx, it is likely that we are erroneously tracing into a custom kernel. To fix this, please wrap the custom kernel into an opaque custom op. Please see the following for details: https://pytorch.org/tutorials/advanced/custom_ops_landing_page.html
E       If you're using Caffe2, Caffe2 uses a lazy allocation, so you will need to call mutable_data() or raw_mutable_data() to actually allocate memory.

```

### What's changed

- pytorch/src/model.py: Local BiLSTM-CRF with padded LSTM (no packing) and fully tensorized Viterbi (no .item() or Python loops).
- model_utils.py: Switched from bi_lstm_crf.BiRnnCrf to local BiRnnCrf.
- loader.py: Added batch_size, device, unpack_forward_output; kept inputs as long indices.
- requirements.txt: Removed bi_lstm_crf dependency.
**- The model currently runs E2E on tt-xla**

### Checklist

### Logs

Before fix:
[bilstm.txt](https://github.com/user-attachments/files/25406909/bilstm.txt)

After Fix:
[after_fix.log](https://github.com/user-attachments/files/25406968/after_fix.log)


ppadjinTT (commented):
Looking at the issue you are solving, I can't see in the logs which part of the model caused the error `RuntimeError: The tensor has a non-zero number of elements, but its data is not allocated yet....`. This PR basically rewrites the whole model and the error message you are getting is from torch, not the compiler.

Can you do the following: Run the test that caused this error and create a minimal reproduction of this error. This seems to be caused by one op/module, not the entire model. Then we can discuss what is the right way to approach this. I am skeptical about needing a whole model rewrite.

@vkovinicTT wdyt?

nvukobratTT (commented):
This is quite a big workaround to avoid torch.nn.LSTM. For now, I'm rather in favor of putting this one on hold until we determine how to handle this case.

@saiarthiraguram, can you open a follow-up issue that is dedicated to the core problem here? Please assign it to me.

saiarthiraguram:
> This is quite a big workaround to avoid torch.nn.LSTM. For now, I'm rather in favor of putting this one on hold until we determine how to handle this case.
> 
> @saiarthiraguram, can you open a follow-up issue that is dedicated to the core problem here? Please assign it to me.

Sure @nvukobratTT

saiarthiraguram:
This issue has been fixed without re-writting the model. Hence closing this PR

PR #523, opened 2026-03-13
Changing transfuser loader

ajakovljevicTT:
Moving the postprocessing to be used just for demos, and tt-xla infra to use the pure transfuser model, as postprocessing cannot currently be compiled through tt-mlir due to `torch.linalg` ops present there. CC @mstojkovicTT @vkovinicTT


vkovinicTT:
> > Should we update how the model is validated, since the output of the model is different now?
> 
> @vkovinicTT The outputs are just compared via pcc through tt-xla infra, same as before. As I do not have the full context of the model, do we need to change that?

If we are only running the pure transfuser model in `tt-xla` infra, then I think everything is fine.

PR #528, opened 2026-03-16
Fix PI0 model: per-device action queues, cumsum bool dtype handling, and deterministic noise for CPU/TT parity

ashokkumarkannan1:
### Problem description

- The PI0 model's custom forward method used a single `_action_queue` (deque) attached to self. The test harness calls forward on the same model instance twice -- once on CPU for the reference output, then once on the TT device for comparison. Since a deque is a plain Python object (not a tensor), it is never copied or moved when switching devices -- both runs shared the exact same queue in memory.

- The CPU run called `sample_actions`, filled the queue with 10 actions, and popped action[0]. The TT run found 9 leftover actions, skipped sample_actions entirely, and popped action[1] from the stale CPU-computed queue. PCC compared completely unrelated values, resulting in `pcc=0.79.`

- After fixing the queue issue, the TT run correctly reached sample_actions but hit a `RuntimeError` on `torch.cumsum` with bool inputs -- eager PyTorch implicitly promotes bool → int64, but the TT-XLA compiler preserves the bool type, causing a `stablehlo.reduce_window legalization failure` during SHLO-to-TTIR conversion.
- Additionally, CPU and TT runs were each generating their own random noise for the diffusion sampling, making PCC comparison non-deterministic.

### What's changed
- Per-device action queues: Replaced the single shared `_action_queue` with a `_device_queues` dictionary keyed by str(state.device). Each device ("cpu", "xla:0") gets its own independent deque, preserving the original queue-drain behavior while ensuring CPU and TT runs never interfere.

- Cumsum bool dtype fix: Added a try/finally patch around `sample_actions` that wraps `torch.cumsum` to auto-cast bool inputs to int64 before calling the original. This replicates what PyTorch already does implicitly -- the cast is lossless (bool 0/1 maps exactly to int64 0/1) and produces bit-identical results. The patch is scoped to sample_actions only and reverts immediately after. Verified with a standalone [test_cumsum_sanity.py](https://github.com/user-attachments/files/26048300/test_cumsum_sanity.py) that checks PCC (1.0), allclose, and output dtype consistency for all 4 cumsum configurations observed in the PI0 inference path. (log for reference - [cumsum_sanity.log](https://github.com/user-attachments/files/26048317/cumsum_sanity.log))

- Deterministic noise: Pre-generate the diffusion noise tensor in `load_inputs` and pass it through forward → sample_actions so both CPU and TT runs start from the same noise, ensuring reproducible PCC comparison.

### Checklist

### Logs
- Before fix - [pi_0_base_before_fix.log](https://github.com/user-attachments/files/26025118/pi_0_base_before_fix.log), [pi_0_libero_before_fix.log](https://github.com/user-attachments/files/26025119/pi_0_libero_before_fix.log)

- After this fix, the model now correctly reaches sample_actions on the TT device. and pi_0_base variant [passes](https://github.com/user-attachments/files/26048335/pi_0_base.log) and pi_0_libero_variant is failing with [PCC Drop of 0.81](https://github.com/user-attachments/files/26048344/pi_0_libero.log) (tracked here - https://github.com/tenstorrent/tt-xla/issues/3786)

### DIFF
--- a/falcon/bounty_jax/loader.py
+++ b/falcon/bounty_jax/loader.py
@@ -1,0 +1,162 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+from typing import Any, Optional
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+from flax import linen
+
+from ...base import ForgeModel
+from ...config import (
+    Framework,
+    ModelConfig,
+    ModelGroup,
+    ModelInfo,
+    ModelSource,
+    ModelTask,
+    StrEnum,
+)
+from transformers import LlamaConfig  # Original used config from LlamaConfig
+from .src.model import FlaxFalcon3ForCausalLMModule
+
+
+class _FalconWrapper(linen.Module):
+    config: Any
+    dtype: Any
+
+    def setup(self):
+        self.llm = FlaxFalcon3ForCausalLMModule(config=self.config, dtype=self.dtype)
+
+    def __call__(self, inputs):
+        input_ids, attention_mask, position_ids = inputs
+        return self.llm(input_ids, attention_mask, position_ids)
+
+
+class ModelVariant(StrEnum):
+
+    CUSTOM_1X2 = "Custom_1x2"
+    CUSTOM_1X4 = "Custom_1x4"
+    CUSTOM_1X8 = "Custom_1x8"
+
+
+class ModelLoader(ForgeModel):
+
+    _VARIANTS = {
+        ModelVariant.CUSTOM_1X2: ModelConfig(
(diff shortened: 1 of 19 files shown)


pi_0/pytorch/src/model.py
```diff
@@ -54,46 +54,67 @@ def forward(
...
+    if len(queue) == 0:
+        original_cumsum = torch.cumsum
+
+        def _safe_cumsum(input, dim, **kwargs):
+            if input.dtype == torch.bool:
+                input = input.to(torch.long)
+            return original_cumsum(input, dim, **kwargs)
+
+        torch.cumsum = _safe_cumsum
+        try:
```
ppadjinTT: i think we should avoid try blocks whenever possible, we should use if's to make sure this works with various inputs

PR #574, opened 2026-04-10
Fix Runtime Error in Gemma3 Model

### LINKED ISSUE #4161: Value out of range error in gemma3
`gemma3/multimodal/pytorch-google/gemma-3-4b-it-single_device-inference` fails with `Value out of range (expected to be in range of [-277, 276], but got -1023)`

meenakshiramanathan1:
### Ticket
Fixes [#4161](https://github.com/tenstorrent/tt-xla/issues/4161)

### Problem description

- `gemma3/multimodal/pytorch-google/gemma-3-4b-it-single_device-inference` fails with `Value out of range (expected to be in range of [-277, 276], but got -1023)
`
### What's changed

- The failure happens because Hugging Face’s sliding-window KV cache assumes the sequence length is at least as large as the configured window. In this case, after concatenating past and current key/value tensors, the total sequence length is only 277, but the cache still slices using a fixed negative index based on a window of 1024 (i.e., -1023). That index is out of bounds for such a short sequence, causing the runtime error. The patch modifies how DynamicSlidingWindowLayer.update trims its cached key and value tensors. Instead of always slicing with a fixed negative index based on the full sliding window size , it first concatenates past and current states into full_key_states and full_value_states, then checks their actual sequence length. It defines max_keep = sliding_window - 1 and only applies the tail slice if the sequence length exceeds this limit and  it keeps the entire tensor unchanged. This avoids invalid negative indexing when the sequence is still shorter than the window.

- There is no changes in final values after these changes in cpu run and model fails with out of memory error post this change.

### Logs
[gemma3_before_fix.log](https://github.com/user-attachments/files/26665927/gemma3_before_fix.log)
[gemma3_modified_latest.log](https://github.com/user-attachments/files/26919861/gemma3_modified_latest.log)

### DIFF
--- a/gemma3/multimodal/pytorch/loader.py
+++ b/gemma3/multimodal/pytorch/loader.py
@@ -7,6 +7,7 @@ Gemma3 model loader implementation for multimodal modeling.
 
 from typing import Optional, Any
 
+import torch
 from transformers import (
     AutoProcessor,
     Gemma3ForConditionalGeneration,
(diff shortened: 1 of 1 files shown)


saiarthiraguram (approved):

meenakshiramanathan1:
Closing this since sliding attention support is [added](https://github.com/tenstorrent/tt-xla/pull/3688) in tt-xla.

PR #582, opened 2026-04-14
Add kimi k2 model (modified)

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3763

### Problem description
We want to add the Kimi-k2 model (modified) to tt-forge-models so that it can be run through the benchmark tester in tt-xla. The modifications to the model are necessary to run this model through tt-xla but all modifications have been marked with `# Modified: <reason>` and the variant is named `KIMI_K2_INSTRUCT_MODIFIED` to avoid confusion.

### What's changed
Added folder kimi_k2 with loader.py, config.json, configuration_deepseek.py and modified_modeling_deepseek.py. 

### Checklist


AleksKnezevic (approved):

kmabeeTT (approved):
Nice, good to document modifications, thx for that. A bit worried about how it will be served through VLLM eventually without these modifications, maybe VLLM's implementation of the model avoids the issues we're seeing here - will have to check.

AleksKnezevic:
vLLM should be much easier since we'll already have the model separated out to us in blocks

gengelageTT:
Benchmark test [passing here](https://github.com/tenstorrent/tt-xla/actions/runs/24426333448) and torch kimi-k2 tests in tt-xla passing locally.

PR #587, opened 2026-04-15
Fix GR00T model for transformers 5.x compatibility

ashokkumarkannan1:
### Ticket
- https://github.com/tenstorrent/tt-xla/issues/4270

### Problem description
-  The GR00T model (issac_groot/pytorch) was skipped due to [#2817](https://github.com/tenstorrent/tt-xla/issues/2817) (slice op issue). On latest main, it now fails with an ImportError because the model source code supported [transformers==4.57.1](https://github.com/tenstorrent/tt-forge-models/pull/403) APIs (DefaultFastImageProcessorKwargs, group_images_by_shape, reorder_images) which were removed/relocated in transformers 5.x.

- `Eagle2_5_VLImageProcessorFast.preprocess()` receives a new `image_seq_length` kwarg that the custom implementation doesn't expect, causing TypeError.

- `GR00T_N1_5.__init__` skips `post_init()`, which means from_pretrained() silently fails to set up tied-weight metadata required by transformers 5.x.

- `FlowmatchingActionHead.__init__ `creates a Beta distribution whose `validate_args=True` default calls .item() on meta tensors during from_pretrained, raising RuntimeError.

- Transformers 5.x changed Qwen2Attention/Qwen3Attention forward to use `.view(*, -1, head_dim)` which is ambiguous for 0-length sequences that the GR00T model legitimately produces during vision-only inference.

- `loader.load_inputs()` didn't mark inputs as preprocessed, causing redundant preprocess() on XLA tensors during forward().

### What's changed

- All fixes are confined to the GR00T model files (`issac_groot/)` and do not alter the model's numerical behavior — only API compatibility and runtime safety.
- Import updates: Replace removed `DefaultFastImageProcessorKwargs` with ImagesKwargs (from transformers.processing_utils); move group_images_by_shape / reorder_images to transformers.image_transforms.
- `Eagle2_5_VLFastImageProcessorKwargs`: Inherit from ImagesKwargs instead of DefaultFastImageProcessorKwargs.
- `preprocess()` kwargs: Pop the new image_seq_length kwarg that transformers 5.x injects.
- `GR00T_N1_5.__init__:` Call post_init() with init_weights temporarily stubbed out, so from_pretrained metadata is set without reinitializing separately-loaded sub-module weights.
- `EagleBackbone.__init__`: Apply instance-scoped `_patch_0len_attention()` to the eagle model's attention layers.
issac_groot/pytorch/src/utils.py
- `_patch_0len_attention`(model): New function that patches Qwen2/Qwen3 attention instances (not the global class) to short-circuit for 0-length sequences, returning zero tensors. The patch is bound per-instance and garbage-collected with the model.
- `FlowmatchingActionHead.__init__`: Pass validate_args=False to Beta() to avoid .item() on meta tensors during from_pretrained.
- `load_inputs()`: Set observations["preprocessed"] = True so forward() skips redundant preprocessing on XLA tensors.

### Checklist
- [x] Verified the changes

#### Note
After the transformers 5.x uplift, CPU-level PCC between the old (4.57.1) and new (5.2.0) outputs dropped to ~0.95. Since none of the changes above touch model weights or computation logic, this drop is attributed to internal implementation changes within the transformers library itself (e.g., attention, image preprocessing). The upstream NVIDIA Isaac-GR00T [source](https://github.com/NVIDIA/Isaac-GR00T/blob/main/pyproject.toml#L28) also still pins `transformers==4.51.3` and has not uplifted to 5.x. Will bisect it later.

### Logs
- [groot.log](https://github.com/user-attachments/files/26759080/groot.log)
- [groot_after_fix.log](https://github.com/user-attachments/files/26759079/groot_after_fix.log) fails with L1 overflow issue that is tracked here - https://github.com/tenstorrent/tt-xla/issues/2817#issuecomment-4254532338


kamalrajkannan78 (approved):

ppadjinTT (approved):
tnx!

PR #598, opened 2026-04-20
[GLiNER] XLA-compatible BiLSTM and static-shape pre/post-processing

### LINKED ISSUE #4302: Gliner model Fails with tensor data not allocated
Root cause: LstmSeq2SeqEncoder.forward called pack_padded_sequence, which internally calls aten._local_scalar_dense to read sequence lengths from XLA lazy tensors at trace time. XLA tensors have no backing memory at trace time.

### Repro

pytest tests/runner/test_models.py::test_all_models_torch[gliner/pytorch-Multi_v2.1-single_device-inference] -svv

### Logs

[gliner.txt](https://github.com/user-attachments/files/26884021/gliner.txt)

Fix: Monkey-patched LstmSeq2SeqEncoder.forward with a padded LSTM that skips packing:

 ```
 def _padded_lstm_forward(self, x, mask, hidden=None):
      output, _ = self.lstm(x, hidden)
      return output * mask.unsqueeze(-1).to(output.dtype)
```

saiarthiraguram:
### Ticket
Fixes: [Github issue](https://github.com/tenstorrent/tt-xla/issues/4302)

### Problem description
Fails with tensor data not allocated

### What's changed
Replace pack_padded_sequence + nn.LSTM (which triggers aten._local_scalar_dense on XLA lazy tensors) with an unrolled mask-gated BiLSTM. The backward direction carries h=0 through tail-padding positions so the first real token from the right is processed from zero hidden state, matching packed-sequence behavior exactly. A naive padded LSTM accumulates bias-driven updates across (seq_len-N) padding positions before reaching the last real token, polluting all backward hidden states.

Also replace extract_word_embeddings and extract_prompt_features with static one-hot selector + einsum variants that avoid torch.where() + data-dependent scatter — the original ops also trigger _local_scalar_dense on X

### Checklist

### Current status
Model fails with PCC issue 
`AssertionError: Evaluation result 0 failed: PCC comparison failed. Calculated: pcc=-0.008028420505954936. Required: pcc=0.99.`
### Logs
before fix: 
[gliner.txt](https://github.com/user-attachments/files/27123046/gliner.txt)

After fix: 
[gliner_after_fix.txt](https://github.com/user-attachments/files/27123050/gliner_after_fix.txt)


vkovinicTT:
Since this is a bit of a bigger change and is kind of hard to follow, please test original implementation vs patched one (both on cpu) and measure whether they are the same. Then upload the logs. That way we can be sure that math checks out and then I will take a closer look at the PR.

saiarthiraguram:
> Since this is a bit of a bigger change and is kind of hard to follow, please test original implementation vs patched one (both on cpu) and measure whether they are the same. Then upload the logs. That way we can be sure that math checks out and then I will take a closer look at the PR.

Here are the logs and the codes tested for numeric equivalence
[test_cpu_equivalence.log](https://github.com/user-attachments/files/27622035/test_cpu_equivalence.log)
[test_cpu_equivalence.py](https://github.com/user-attachments/files/27622036/test_cpu_equivalence.py)

vkovinicTT (approved):
I am approving this, just open an issue which will explain why we can't run BiLSTM as is with detailed explanation.

saiarthiraguram:
> I am approving this, just open an issue which will explain why we can't run BiLSTM as is with detailed explanation.

https://github.com/tenstorrent/tt-xla/issues/4302 - attached the details here

PR #605, opened 2026-04-21
Add LoRA versions of models

agobeljicTT:
### Ticket
Part of https://github.com/tenstorrent/tt-xla/issues/4107
> [!NOTE]
> Stacked with https://github.com/tenstorrent/tt-xla/pull/4219/

### Problem description                                                                                                                        
We need LoRA-adapted variants of our causal LM models to test fine-tuning                                                                     on Tenstorrent hardware via tt-xla.

### What's changed                                                                                                                             

**New shared infrastructure (`tools/`)**                                                                                                       
- `tools/lora.py` - Replaces the old standalone `apply_lora_adapters` function with `LoRAModelMixin`. The mixin overrides `load_model` via Python MRO, wrapping the base model with a PEFT LoRA adapter declared in the subclass's `_LORA_CONFIGS` dict. It also appends " LoRA" to `ModelInfo.model` so registry lookups distinguish LoRA from base variants.                                                                   
- `tools/prefill_inputs.py` - Promotes `llama/causal_lm/pytorch/prefill_inputs.py` to a shared tool and adds `PrefillInputsMixin`, making `load_inputs_prefill` reusable across all causal-LM loaders without copy/paste.

**New LoRA model loaders** (thin subclasses via `LoRAModelMixin`):
- `llama_lora` - TinyLlama-v1.1, Llama-3.2-1B (r=4), Llama-3.2-3B (r=16)
- `gemma_lora` - Gemma-1.1-2B-IT (r=4)
- `phi1_lora` - Phi-1 (r=4)
- `qwen_2_5_lora` - Qwen-2.5-1.5B (r=4)
                                                                                                                                                 
All LoRA configs target `["q_proj", "v_proj"]` attention projections.                                                                          
                                                                                                                                                 
**Base model updates**                                                                                                                         
- `llama`, `gemma`, `phi1`, `qwen_2_5` loaders now inherit `PrefillInputsMixin` (removing duplicate `load_inputs_prefill` implementations).
- `load_shard_spec` in all four base loaders now calls `model.get_base_model()` when available, so sharding works correctly when a PEFT wrapper is passed in.
- `phi1` loader gains `load_inputs_decode` and `load_config` to reach parity with the other causal-LM loaders.

### Checklist


phi1/causal_lm/pytorch/loader.py
```diff
@@ -130,8 +137,35 @@ def load_model(self, *, dtype_override=None, **kwargs):
 
         model = PhiForCausalLM.from_pretrained(pretrained_model_name, **model_kwargs)
 
+        self.model = model
+        self.config = model.config
+
         return model
 
+    def load_config(self):
```
agobeljicTT, line 135: perhaps not part of this PR... but still necessary

tools/prefill_inputs.py
```diff
@@ -769,3 +773,59 @@ def get_prefill_texts_for_batch(seq_len: int, batch_size: int) -> list:
...
+            padding="max_length",
+            truncation=True,
+            max_length=seq_len,
+        )
+
+        if dtype_override is not None:
+            for key in inputs:
+                inputs[key] = cast_input_to_type(inputs[key], dtype_override)
+
+        self.seq_len = seq_len
```
agobeljicTT: this probably should be wrapped in a getter/setter fashion but i dont think it's mandatory, curious to what you think

agobeljicTT: Looked around, there might be need to guard this despite claude saying no:
```
These three callers are three separate use contexts:                                                                                                                                                                        
                                                                                                                                                                                                                                              
  1. _get_input_activations (test harness) — calls dynamic_loader.load_inputs(run_phase=...) which routes to load_inputs, load_inputs_decode, or load_inputs_prefill depending on the phase. This is the main test execution path.            
  2. load_parameters / load_parameters_partition_spec (JAX bounty loaders like mistral, mixtral) — calls load_inputs directly to get a concrete input shape so they can trace parameter partition specs. They need real tensors to figure out 
  sharding, not to run inference.                                                                                                                                                                                                             
  3. decode_output (albert, phi1, etc.) — calls load_inputs as a fallback when inputs=None. It's a convenience: if you call decode_output standalone without feeding it inputs, it generates its own. Lazy self-contained default.            
                                                                                                                                                                                                                                              
  So load_inputs serves three different roles that happen to share the same method:                                                                                                                                                           
  - Test entry point for non-phase-aware models                                                                                                                                                                                               
  - Shape probe for parameter initialization                                                                                                                                                                                                  
  - Lazy default for output decoding helpers                           
                                                                                                                                                                                                                                              
  They're unrelated concerns that all converged on load_inputs as the "get me some inputs" primitive.
```

I'll see either in this PR or in another

agobeljicTT: made a queued PR: https://github.com/tenstorrent/tt-forge-models/pull/616

tools/lora.py
```diff
@@ -1,43 +1,51 @@
...
-        target_modules: Names of linear submodules to apply LoRA to.
-            Defaults to ``["q_proj", "v_proj"]``, which covers Llama, GLM,
-            and Qwen attention projections.
-
-    Returns:
-        A ``peft.PeftModel`` wrapping the original model.
+class LoRAModelMixin:
+    """Mixin that adds LoRA adapter wrapping to any ForgeModel subclass.
+
+    Subclasses must define ``_LORA_CONFIGS`` mapping variants to
```
ppadjinTT, line 13: not sure if it is the best practice to say in one docstring that classes should implement a private class field. If you already picked to make a base class, you should probably request this behavior (since comments/docstrings are less and less trustworthy with agents).

Consider making `LoRAModelMixin` a base class that needs to implement get_lora_configs abstract method

ppadjinTT: or i guess since you already have _DEFAULT_LORA_CONFIG, you can make this a python property lora_configs and have the default behavior to return _DEFAULT_LORA_CONFIG.

base.py
```diff
@@ -271,16 +274,59 @@ class ForgePrefillModel(ForgeModel):
...
+
+        Args:
+            dtype_override: Optional torch.dtype to override the model's default dtype.
+            batch_size: Batch size for the inputs.
+            seq_len: Target sequence length. Texts are chosen to minimize padding.
+
+        Returns:
+            dict: Input tensors (input_ids, attention_mask) padded to seq_len.
+        """
+        assert hasattr(
```
ppadjinTT: i think these should be required by the ForgePrefillModel base class when you inherit, i don't see the benefit of it being in the runtime

ppadjinTT (approved):
in general good, i gave you few OOP-related comments. You don't have to address them, but it will probably help reduce the chance of introducing bugs

tools/lora.py
```diff
@@ -1,43 +1,51 @@
...
+    ``peft.LoraConfig``. Variants not listed fall back to
+    ``_DEFAULT_LORA_CONFIG``.
+
+    Must be listed before the base model loader in the class bases so that
+    Python's MRO resolves ``load_model`` here first.
```
vkovinicTT, line 18: Similar comment to @ppadjinTT 's comment, we write here that we need to inherit this before other class like [here](https://github.com/tenstorrent/tt-forge-models/pull/605/changes#diff-8e1b1028ddcc2b37df8a744eeeca013a83743da9871af8d43315c1b89f80a175R27) for example, but we don't have any guarantees that someone will. Can we somehow enforce it?

agobeljicTT: `ClassName.mro()[1]` or `ClassName.__mro__`
But this is too hacky

tools/lora.py
agobeljicTT: the idea was that this is just an interface that has a hidden state
on the other hand, we can use a function that returns a dict, one can just override it `get_lora_configs`

base.py
agobeljicTT: stale check will remove

gemma/pytorch/loader.py
```diff
@@ -206,7 +206,12 @@ def get_mesh_config(self, num_devices: int):
             ), "Attention heads must be divisible by the model axis size"
         return mesh_shape, ("batch", "model")
 
-    def load_shard_spec(self, model):
+    def load_shard_spec(self, model, strategy="fsdp", batch_axis="batch"):
```
umalesTT: For basic `ModelLoader` we should stick to backward-compatibility, i.e. keep existing signature of `load_shard_spec` - if we want more extensive testing in terms of sharding, it should be implemented (as it is) below in `ModelLoaderPrefill`

gemma/pytorch/loader.py
```diff
@@ -258,3 +263,59 @@ def load_inputs_decode(self, dtype_override=None, batch_size=1):
...
+    DEFAULT_VARIANT = ModelVariant.GEMMA_1_1_2B_IT
+
+    def load_shard_spec(self, model, strategy="fsdp", batch_axis="batch"):
+        """Weight shard spec parameterized by ``strategy`` and ``batch_axis``
+        (use "data" when inputs are also sharded).
+        """
+        shard_specs = {}
+
+        if strategy == "fsdp":
+            shard_specs[model.model.embed_tokens.weight] = (None, batch_axis)
```
umalesTT: Did you check for weights names of Gemma 1.1 2b? Do they all match with shardings defined here?

llama/causal_lm/pytorch/loader.py
```diff
@@ -365,11 +364,12 @@ def get_mesh_config(self, num_devices: int):
 
         return mesh_shape, ("batch", "model")
 
-    def load_shard_spec(self, model):
+    def load_shard_spec(self, model, strategy="fsdp", batch_axis="batch"):
```
umalesTT: Same comment as above

phi1/causal_lm/pytorch/loader.py
```diff
@@ -169,3 +208,76 @@ def decode_output(self, outputs, inputs=None):
...
+                shard_specs[layer.self_attn.dense.weight] = (batch_axis, "model")
+                shard_specs[layer.self_attn.dense.bias] = (batch_axis,)
+                shard_specs[layer.input_layernorm.weight] = (batch_axis,)
+                shard_specs[layer.input_layernorm.bias] = (batch_axis,)
+
+        elif strategy == "megatron":
+            shard_specs[model.model.embed_tokens.weight] = (None, None)
+            shard_specs[model.lm_head.weight] = ("model", None)
+            shard_specs[model.lm_head.bias] = ("model",)
+            shard_specs[model.model.final_layernorm.weight] = (None,)
```
umalesTT: You don't need explicit replication with `None`, if not specified, weight will be replicated

phi1/causal_lm/pytorch/loader.py
```diff
@@ -118,8 +124,41 @@ def load_model(self, *, dtype_override=None, **kwargs):
...
         return model
 
+    def load_config(self):
+        """Load and return the configuration for the Phi-1 model variant."""
+        self.config = AutoConfig.from_pretrained(
+            self._variant_config.pretrained_model_name
+        )
+        return self.config
+
+    def load_shard_spec(self, model, strategy="fsdp", batch_axis="batch"):
```
umalesTT: Same as above

umalesTT (approved):

PR #610, opened 2026-04-23
Add clean model names to model_info json

vvukomanTT:
### Ticket
/

### Problem description
Customer team needs clean model names so they can consolidate data model coverage data from multiple tenstorrent repos.

### What's changed
Add a new field to the model test data that's uploaded to superset.
The field contains a clean model name.

### Checklist

### DIFF
--- a/base.py
+++ b/base.py
@@ -116,7 +116,7 @@ class ForgeModel(ABC):
     def get_model_info(cls, variant: Optional[StrEnum] = None) -> ModelInfo:
         """Get model information for dashboard and metrics reporting.
 
-        Automatically populates model_name_hf from the variant's pretrained_model_name
+        Automatically populates model_name_clean from the variant's pretrained_model_name
         when the loader uses the _VARIANTS system.
 
         Args:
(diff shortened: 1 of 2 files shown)


base.py
```diff
@@ -115,16 +116,23 @@ def get_variant_config(
...
                      If None, DEFAULT_VARIANT is used.
 
-
         Returns:
             ModelInfo: Information about the model and variant
         """
         variant_enum = cls._validate_variant(variant)
-        return cls._get_model_info(variant_enum)
+        info = cls._get_model_info(variant_enum)
+        if info.model_name_hf is None:
```
vkovinicTT: We want to add HF name only if it is Hugging Face model, right?
I think something like this would be more fitting:
```python
if info.model_name_hf is None and info.source == ModelSource.HUGGING_FACE:
```

vkovinicTT (approved):

base.py
vvukomanTT: I've made the field name more general, since the `org/name` pattern formatting can work for other models as well 👍

PR #612, opened 2026-04-23
Introduce ForgePrefillModel class for extensive prefill testing

umalesTT:
### Ticket
None

### Problem description
Prefill-specific functions currently live on the regular `ModelLoader`, even though only a few variants are swept across many sequence lengths and batch sizes. We want better modularity: a clear separation between the regular path and the prefill-focused path, so each lives in its own place and is easier to extend.

### What's changed
Added a `ForgePrefillModel` base class in `base.py` that holds the prefill-specific functions in one place, designed for variants tested across many sequence lengths, batch sizes and sharding strategies. Llama's loader is split so the regular path and the prefill path live in separate classes, and the shared prefill text data (covering the supported sequence lengths and batch sizes) is moved to a top-level module so other prefill loaders can reuse it later.
Relevant `tt-xla` [PR](https://github.com/tenstorrent/tt-xla/pull/4390)

### Checklist


llama/causal_lm/pytorch/loader.py
```diff
@@ -331,46 +331,6 @@ def load_inputs_decode(self, dtype_override=None, batch_size=1):
             dtype=dtype_override,
         )
 
-    def load_inputs_prefill(self, dtype_override=None, batch_size=1, seq_len=128):
```
ppadjinTT, line 319: when we discussed offline, you mentioned that this function might get used for non-prefill tests. maybe this could have a different name? `load_batched_inputs` seems more general perhaps

llama/causal_lm/pytorch/loader.py
```diff
@@ -420,21 +380,11 @@ def get_mesh_config(self, num_devices: int):
...
-                        sharding axis, forcing the mesh to ("data", "model").
+    def load_shard_spec(self, model):
+        """Default shard spec on a ("batch", "model") mesh.
 
-        Returns:
-            dict mapping weight tensors to shard spec tuples, or None for small models.
+        Used by non-prefill TP paths; see :class:`ModelLoaderFocus` for the
+        strategy-parameterized version.
         """
         if self._variant in [
```
ppadjinTT, line 374: maybe add some log.info saying that model sharding is not applied to <model_that_should_skip_sharding>. otherwise user would assume that his model will get sharded. 

Also consider renaming function to `maybe_load_shard_spec`

base.py
```diff
@@ -262,3 +262,26 @@ def get_weight_dtype_config_path(self) -> Optional[str]:
                 f"Available configs: {os.listdir(config_dir)}"
             )
         return None
+
+
+class ForgeFocusModel(ForgeModel):
```
ppadjinTT: as we discussed offline, this name might need some change. wdyt @vzeljkovicTT?  fyi @agobeljicTT

ppadjinTT (commented):
Changes are good, I'm not approving yet to leave some time for others to see the naming discussions.

base.py
agobeljicTT: I am 100% for the name change, I'm currently reviewing both PRs and will come up with something...

agobeljicTT: my take is that prefill logic should be independent, so we can make a PrefillMixin or something (#605 for reference) that will group everything and be included when necessary

prefill_inputs.py
agobeljicTT, line 1: i'd move this to `tools/prefill_inputs.py` rather then root

agobeljicTT (commented):
I suggest we have an offline discussion.
Here are Claude's thoughts:
```
They solve different problems and shouldn't be framed as either/or:                                                                                                                 
  
  - #605 is the broader, more reusable foundation. Mixins (PrefillInputsMixin, LoRAModelMixin) are the right tool when many loaders share the same plumbing — putting prefill in      
  tools/ and composing it via MRO eliminates copy/paste across 4+ model families. The PEFT-unwrap in load_shard_spec is the real architectural win: it makes LoRA-wrapped models
  transparent to any sharding caller.                                                                                                                                                 
  - #612 is the right shape for a narrower concern — flagging which loaders are intended for the prefill sweep matrix (mesh × strategy × batch × seq_len). The strategy-parameterized
  load_shard_spec (fsdp / megatron) is genuinely focus-specific and doesn't belong on every loader.                                                                                   
  
  How they complement each other                                                                                                                                                      
                            
  A clean merged design would keep both ideas, but consolidate them:                                                                                                                  
                            
  1. Use #605's tools/prefill_inputs.py + PrefillInputsMixin as the implementation of prefill loading (shared, mixin-based).                                                          
  2. Use #612's ForgeFocusModel ABC purely as a marker contract — not as the place that holds the prefill code. ForgeFocusModel would declare the focus-specific surface
  (load_shard_spec(model, strategy, batch_axis)) and rely on PrefillInputsMixin to satisfy load_inputs_prefill.                                                                       
  3. ModelLoaderFocus(PrefillInputsMixin, ModelLoader, ForgeFocusModel) — gets shared prefill from the mixin, focus-specific shard sweep from the subclass, and registers as a focus
  model via the ABC.                                                                                                                                                                  
  4. For LoRA + focus combinations (likely future need), LoRAModelMixin composes orthogonally — class LlamaLoRAFocus(LoRAModelMixin, PrefillInputsMixin, ModelLoader, 
  ForgeFocusModel). The PEFT-unwrap in LoRAModelMixin.load_shard_spec already handles the super() chain correctly.                                                                    
                            
  Practical sequencing recommendation                                                                                                                                                 
                            
  - Land #605 first (it's broader, reorganizes tools/, and the test/contract surface is more stable — no ABC to align on yet).                                                        
  - Rebase #612 on top: drop the prefill move (already done by #605), drop load_inputs_prefill from ModelLoaderFocus (inherit it from PrefillInputsMixin), and keep just the
  ForgeFocusModel ABC + the strategy-parameterized load_shard_spec + the variant restriction. That makes #612 a much smaller, sharper PR — exactly the contract change it advertises, 
  with no rename collateral.
```

base.py
ppadjinTT: btw what is "mixin" in this context

agobeljicTT: @ppadjinTT it's like an interface
@umalesTT synced offline, basically this PR makes sense, just the name xd

llama/causal_lm/pytorch/loader.py
agobeljicTT: there are already three options: load_inputs, load_prefill and load_decode.
this requires a more thorough check

llama/causal_lm/pytorch/loader.py
```diff
@@ -404,21 +365,11 @@ def get_mesh_config(self, num_devices: int):
...
-        Args:
-            model: The model whose weights are to be sharded.
-            strategy: Sharding strategy — "fsdp" shards across both axes,
-                      "megatron" shards on "model" axis only (other axis is None).
-            batch_axis: Name of the non-model mesh axis for "fsdp" specs (ignored
-                        by "megatron"). Defaults to "batch" for a ("batch", "model")
-                        mesh; pass "data" when input sharding is enabled, because
-                        load_shard_spec_data_parallel hardcodes "data" as the input
-                        sharding axis, forcing the mesh to ("data", "model").
+    def load_shard_spec(self, model):
```
agobeljicTT, line 368:
```suggestion
    def load_shard_spec(self, model, **kwargs):
```

wdythink?

Feature | How it fixes MRO
-- | --
*args | Captures all positional arguments to pass them up.
**kwargs | Captures all keyword arguments to pass them up.
Defensive Programming | Every method in the chain takes **kwargs even if it doesn't use them.

llama/causal_lm/pytorch/loader.py
```diff
@@ -430,10 +381,87 @@ def load_shard_spec(self, model, strategy="fsdp", batch_axis="batch"):
...
+    extensively with various meshes, strategies, batches and sequence lengths.
+    """
+
+    _VARIANTS = {
+        ModelVariant.LLAMA_3_2_1B: ModelLoader._VARIANTS[ModelVariant.LLAMA_3_2_1B],
+        ModelVariant.LLAMA_3_1_8B_INSTRUCT: ModelLoader._VARIANTS[
+            ModelVariant.LLAMA_3_1_8B_INSTRUCT
+        ],
+    }
+    DEFAULT_VARIANT = ModelVariant.LLAMA_3_2_1B
```
agobeljicTT, line 426:
```suggestion
    _FOCUS_VARIANTS = (ModelVariant.LLAMA_3_2_1B, ModelVariant.LLAMA_3_1_8B_INSTRUCT)
    _VARIANTS = {v: ModelLoader._VARIANTS[v] for v in _FOCUS_VARIANTS}
    DEFAULT_VARIANT = ModelVariant.LLAMA_3_2_1B
```

agobeljicTT: if they are the same, which i assume they will be

vzeljkovicTT (approved):

PR #616, opened 2026-04-26
Remove seq_len instance state and dead decode generation loops

agobeljicTT:
> [!NOTE]
> This PR is queued on top of #605. The diff shown is relative to that branch.

### Ticket
None

### Problem description
`self.seq_len` was used as implicit shared state between `load_inputs*` methods and `decode_output` — calling `load_inputs` silently set an instance attribute that `decode_output` then read as its generation cursor. This temporal coupling meant any new model adding a LoRA variant also had to manually unwrap the PEFT wrapper in `load_shard_spec` with the same inline ternary, and `PrefillInputsMixin` required host classes to declare a `self.seq_len` field it would write to as a side effect.

### What's changed
- **`self.seq_len` removed entirely** from all model loaders (`llama`, `glm`, `gemma`, `qwen_2_5`, `qwen_3`, `falcon`, `phi1`, `gemma3`, `deepseek_coder`, `deepseek_math`, `gemma/codegemma`). The field was either written and never read (decode models), or served as a side-channel between `load_inputs` and `decode_output`.
- **`PrefillInputsMixin`** no longer writes back `self.seq_len = seq_len` at the end of `load_inputs_prefill` and no longer asserts the attribute exists. The `seq_len` parameter is already known to the caller; if needed it is readable from `inputs["input_ids"].shape[-1]`.
- **Dead `decode_output` generation loops removed** from `llama`, `glm`, `deepseek_coder`, and `deepseek_math`. These autoregressive generation helpers had no callers in any test or framework code. Removing them is more honest than fixing their signature.
- **Orphaned imports cleaned up** (`import torch`, `generate_no_cache`) in `deepseek_coder` and `deepseek_math` that were only used by the removed methods.

### Checklist


nvukobratTT:
Staled PR, closing. Please reopen if still relevant.

PR #628, opened 2026-05-01
Add deepseek-v3.2-exp

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3763

### Problem description
We want to add the Deepseek-v3.2-exp model (modified) to tt-forge-models so that it can be run through the benchmark tester in tt-xla. The modifications to the model are necessary to run this model through tt-xla but all modifications have been marked with # Modified: <reason> and the variant is named DEEPSEEK_V2_3_EXP_MODIFIED to avoid confusion.

### What's changed
Added folder deepseek/deepseek_v3_2_exp with loader.py, config.json, modified_model.py and original_model.py.

In loader.py the transformer is wrapped in a custom `DeepSeekV32ForCausalLM` since the benchmark calls `model(input_ids=..., past_key_values=..., cache_position=..., use_cache=...)` and reads `output.logits`. The transformer in modified_model.py uses `forward(tokens, start_pos)` and returns a raw `[batch, vocab_size]` tensor. This wrapper bridges that gap without touching `modified_model.py` when not necessary. 

### Checklist


gengelageTT:
Performance benchmark passing here: https://github.com/tenstorrent/tt-xla/actions/runs/25188580714

AleksKnezevic (approved):

kmabeeTT (approved):
Thanks for changes, a bit worried about VLLM like previous KimiK2 PR which I'll link here: https://github.com/tenstorrent/tt-forge-models/pull/582#pullrequestreview-4107348408 but but the thought was it was ~lower risk, so we can re-evaluate when the time comes.

PR #634, opened 2026-05-05
Add hunyuan video component loaders

meenakshiramanathan1:
### Ticket

https://github.com/tenstorrent/tt-xla/issues/4519

### Problem description
Add component loaders for the HunyuanVideo pipeline

### What's changed
New package: `hunyuan_video/pytorch/`

| File | Purpose |
|------|---------|
| `loader.py` | ForgeModel subclass with four variants (TextEncoder, TextEncoder2, Transformer, Vae), each loading only its own component via `from_pretrained(subfolder=...)` |
| `src/model_utils.py` | Component loaders, wrapper modules, shard specs, mesh shapes, inference shape constants |

Component sources:

| Variant | Class | 
|---------|-------|
| `TextEncoder` | LLaMA-3 encoder
| `TextEncoder2` | CLIP text encoder
| `Transformer` | HunyuanVideoTransformer3DModel DiT
| `Vae` | AutoencoderKLHunyuanVideo

Each component is loaded independently — no full pipeline is instantiated.

`ModelLoader` API:
- `load_model(dtype_override)` — returns the component as a `torch.nn.Module`
- `load_inputs(dtype_override)` — returns synthetic input tensors matched to the component's forward signature
- `get_mesh_config(num_devices)` — returns `(mesh_shape, mesh_names)` for a 2D `(batch, model)` SPMD mesh 
- `load_shard_spec(model)` — returns per-tensor partition specs for tensor parallelism

### Checklist
- [x] Verify changes by local testing in Wormhole

### Logs

[hunyuan_text_encoder_e2e.log](https://github.com/user-attachments/files/27597536/hunyuan_text_encoder_e2e.log)
[hunyuan_text_enocder2.log](https://github.com/user-attachments/files/27597540/hunyuan_text_enocder2.log)
[hunyuan_transformer.log](https://github.com/user-attachments/files/27597551/hunyuan_transformer.log)
[hunyuan_vae_decoder.log](https://github.com/user-attachments/files/27597527/hunyuan_vae_decoder.log)

### DIFF
--- a/hunyuan_video/pytorch/loader.py
+++ b/hunyuan_video/pytorch/loader.py
@@ -1,0 +1,162 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""
+HunyuanVideo component loader.
+
+Each variant corresponds to one independently loadable component:
+  - TextEncoder   → LLaMA encoder (text_encoder)
+  - TextEncoder2  → CLIP encoder (text_encoder_2)
+  - Transformer   → HunyuanVideoTransformerWrapper (DiT)
+  - Vae           → VAEDecoderWrapper (decoder-only)
+"""
+
+from typing import Optional
+
+import torch
+
+from ...base import ForgeModel
+from ...config import (
+    Framework,
+    ModelConfig,
+    ModelGroup,
+    ModelInfo,
+    ModelSource,
+    ModelTask,
+    StrEnum,
+)
+from .src.model_utils import (
+    DTYPE,
+    REPO_ID,
+    MESH_NAMES,
+    MESH_SHAPES,
+    HunyuanVideoTransformerWrapper,
+    VAEDecoderWrapper,
+    load_text_encoder,
+    load_text_encoder_2,
+    load_text_encoder_2_inputs,
+    load_text_encoder_inputs,
+    load_transformer,
+    load_transformer_inputs,
+    load_vae,
+    load_vae_inputs,
+    shard_text_encoder_2_specs,
+    shard_text_encoder_specs,
+    shard_transformer_specs,
+    shard_vae_specs,
+)
+
+
+class ModelVariant(StrEnum):
+    """Loadable components of the HunyuanVideo pipeline."""
+
+    TEXT_ENCODER = "TextEncoder"
+    TEXT_ENCODER_2 = "TextEncoder2"
+    TRANSFORMER = "Transformer"
+    VAE = "Vae"
+
+
+class ModelLoader(ForgeModel):
+    """Load individual HunyuanVideo components without pulling the full pipeline."""
+
+    _VARIANTS = {
+        ModelVariant.TEXT_ENCODER: ModelConfig(pretrained_model_name=REPO_ID),
+        ModelVariant.TEXT_ENCODER_2: ModelConfig(pretrained_model_name=REPO_ID),
+        ModelVariant.TRANSFORMER: ModelConfig(pretrained_model_name=REPO_ID),
+        ModelVariant.VAE: ModelConfig(pretrained_model_name=REPO_ID),
+    }
+    DEFAULT_VARIANT = ModelVariant.TRANSFORMER
+
+    @classmethod
+    def _get_model_info(cls, variant: Optional[ModelVariant] = None) -> ModelInfo:
+        if variant is None:
+            variant = cls.DEFAULT_VARIANT
+        task = (
+            ModelTask.NLP_EMBED_GEN
+            if variant in (ModelVariant.TEXT_ENCODER, ModelVariant.TEXT_ENCODER_2)
+            else ModelTask.MM_VIDEO_TTT
+        )
+        return ModelInfo(
+            model="HunyuanVideo",
+            variant=variant,
+            group=ModelGroup.RED,
+            task=task,
+            source=ModelSource.HUGGING_FACE,
+            framework=Framework.TORCH,
+        )
+
+    def load_model(self, *, dtype_override: Optional[torch.dtype] = None, **kwargs):
+        """Load and return the component for this variant as a torch.nn.Module.
+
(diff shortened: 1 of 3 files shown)


hunyuan_video/pytorch/loader.py
```diff
@@ -54,11 +51,11 @@
     shard_transformer_specs,
 )
 
-_REPO_ID = "hunyuanvideo-community/HunyuanVideo-1.5-Diffusers-480p_t2v_distilled"
+_REPO_ID = "hunyuanvideo-community/HunyuanVideo"
```
vkovinicTT: Why don't we import `_REPO_ID` from `model_utils.py` as well?

hunyuan_video/pytorch/loader.py
vkovinicTT: Even though `TEXT_ENCODER_2` and `VAE` fit on a single chip, shouldn't we shard them as well (perf gain). Also once we have everything in pipeline, we'd probably want to have everything sharded.

hunyuan_video/pytorch/loader.py
```diff
@@ -151,28 +148,26 @@ def load_shard_spec(self, model):
...
-        TEXT_ENCODER_2 → [input_ids (1,256) int64, attention_mask (1,256) float32]
+        TEXT_ENCODER   → [input_ids (1,256) int64, attention_mask (1,256) int64]
+        TEXT_ENCODER_2 → [input_ids (1,77) int64, attention_mask (1,77) int64]
         TRANSFORMER    → [hidden_states, timestep, encoder_hidden_states,
-                          encoder_attention_mask, encoder_hidden_states_2,
-                          encoder_attention_mask_2, image_embeds]
-        VAE            → [z (1,32,5,30,53) bfloat16]
+                          encoder_attention_mask, pooled_projections]
+        VAE            → [z (1,16,2,16,16) bfloat16]
         """
```
vkovinicTT, line 150: Please write `load_inputs` function similarly to how you wrote `load_model` where each compoment has it's own loader logic inside `model_utils.py`, separated from the main `load_inputs` as it's more readable that way.

hunyuan_video/pytorch/src/model_utils.py
```diff
@@ -156,7 +152,7 @@ def forward(
 
 
 def shard_text_encoder_specs(encoder) -> dict:
```
vkovinicTT, line 219: Please give me more context on how did you choose strategy for sharding t`ext_encoder` and `transformer`? I am gathering info as I hope we could write claude skill for sharding so that anybody can easily shard model even if they don't have much context in sharding.

vkovinicTT (changes requested):

hunyuan_video/pytorch/src/model_utils.py
meenakshiramanathan1: Sure will share a summary with you.

saiarthiraguram (approved):

hunyuan_video/pytorch/src/model_utils.py
```diff
@@ -0,0 +1,490 @@
...
+def shard_vae_specs(vae) -> dict:
+    """Shard specs for AutoencoderKLHunyuanVideo (decoder-only path).
+
+    For Conv3d weights (out_ch, in_ch, T, H, W) the spec alternates:
+      Column-parallel (entry conv, first conv in resnet, upsampler conv):
+        ("model", "batch", None, None, None)
+      Row-parallel   (second conv in resnet, conv_shortcut, exit conv):
+        ("batch", "model", None, None, None)
+    Norm / attention specs mirror the LLaMA / transformer conventions.
+    """
```
vkovinicTT, line 390: Currently for sharding VAE you can use guide I left [here](https://github.com/tenstorrent/tt-xla/discussions/4664). We only want to shard along one axis, not both axis, as we currently can only shard along channel axis in `conv3d`. In your case it would probably work since your mesh is something like `(1,2)` if you are running on n300, so one of the axis will be 1 anyway and that axis is not being sharded (it's replicated) so the sharding would seem like it's working, but if you were to run it on let's say galaxy which has mesh shape something like `(8,4)` your model would probably fail.

vkovinicTT (changes requested):

hunyuan_video/pytorch/src/model_utils.py
meenakshiramanathan1: Modified sharding along single axis according to guide.

hunyuan_video/pytorch/src/model_utils.py
```diff
@@ -0,0 +1,515 @@
...
+    w2, b2 = _causal_conv3d_weight(block.conv2)
+    specs[w2] = (None, "model", None, None, None)
+    if b2 is not None:
+        specs[b2] = ("batch",)
+
+    if getattr(block, "conv_shortcut", None) is not None:
+        ws, bs = _causal_conv3d_weight(block.conv_shortcut)
+        specs[ws] = (None, "model", None, None, None)
+        if bs is not None:
+            specs[bs] = ("batch",)
```
vkovinicTT: You usually don't want to shard `conv_shortcut` as if you have `conv1` -> ... `conv2` (`conv1` is column parallel and `conv2` is row parallel), by the time you get to the `conv2` you will need to perform `all_gather` which means that you will have `conv2` output replicated, so when you do `conv2`'s `output` + `conv_shortcut` you need both of the tensors replicated. For `row_parallel` you don't shard bias (`b2`). Also pay attention whether math checks out when you shard norm layers (because sometimes you need the statistics of the whole tensor instead of just the part of the tensor that is on local chip).

meenakshiramanathan1: `conv_shortcut` is fully replicated now so it lines up with conv2's replicated output for the residual add, and row-parallel biases (b2, conv_out.bias) are no longer sharded.

On the norm thing — went and checked the lowered IR to make sure the math actually works out.

vkovinicTT (approved):

PR #702, opened 2026-05-28
Add Z-Image transformer SPMD shard specs for multichip bring-up.

ashokkumarkannan1:
### Ticket
- Z-Image transformer compile failure (RoPE complex gather): [tt-xla#4756](https://github.com/tenstorrent/tt-xla/issues/4756) — which is fixed in [tt-mlir#8637](https://github.com/tenstorrent/tt-mlir/pull/8637)

### Problem description
- `ZImageTransformer2DModel` (~6.15B params) does not fit on a single device after the RoPE SHLO→TTIR fix in tt-mlir. tt-xla multichip runs need **partition specs and mesh config** on the forge loader so benchmarks/tests can call `get_mesh_config` / `load_shard_spec` and apply `xs.mark_sharding` before `torch.compile`.
- Without this change, the Z-Image transformer loader only supports single-chip bring-up and cannot drive SPMD sharding from the standard model-discovery / benchmark path.

#### **`z_image/pytorch/src/model_utils.py`**
- `MESH_SHAPES` / `MESH_NAMES` for device counts **1, 2, 4, 8, 32** on a `("batch", "model")` mesh (e.g. 32 → `(8, 4)`).
- `shard_transformer_specs()` — Megatron-style specs for `ZImageTransformer2DModel`:
  - Column-parallel Q/K/V and FFN `w1`/`w3`: `("model", "batch")`
  - Row-parallel O and FFN `w2`: `("batch", "model")`
  - Covers main layers, noise/context refiners, embedders, timestep/caption MLPs, final layers.
  - Pad tokens (`x_pad_token`, `cap_pad_token`, `siglip_pad_token`) fully replicated (rank-2 tensors must not get 1-axis specs).
#### **`z_image/pytorch/loader.py`**
- `get_mesh_config(num_devices)` — transformer uses `MESH_SHAPES`; text encoder / VAE return `(1, 1)` (single-chip components).
- `load_shard_spec(model)` — delegates to `shard_transformer_specs` for `ModelVariant.TRANSFORMER` only.
Text encoder and VAE remain single-chip; only the DiT transformer is sharded.
### Checklist

### Logs
- [zimage_transformer_log_after_fix.log](https://github.com/user-attachments/files/28345602/zimage_transformer_log_after_fix.log)


kamalrajkannan78 (approved):

vkovinicTT (changes requested):
When sharding, please just use just "model" axis and instead of "batch" axis use `None` as true Megatron style sharding shards on only 1 axis

ashokkumarkannan1:
> When sharding, please just use just "model" axis and instead of "batch" axis use `None` as true Megatron style sharding shards on only 1 axis

@vkovinicTT thanks for the review, updated them
- weight shard specs now use Megatron-style single-axis TP (("model", None) / (None, "model")) instead of hybrid ("model", "batch") specs.

vkovinicTT (approved):

PR #717, opened 2026-06-08
Add Gemma2TextEncoderWrapper for Lumina Image Text Encoder

### LINKED ISSUE #4900: [Lumina Image] Text Encoder Runtime Error
`tests/torch/models/lumina_image/test_text_encoder.py::test_text_encoder_sharded` fails with `RuntimeError: Value out of range (expected to be in range of [-256, 255], but got -4095)`

meenakshiramanathan1:
### Ticket

- Fixes [#4900](https://github.com/tenstorrent/tt-xla/issues/4900)

### Problem description

tests/torch/models/lumina_image/test_text_encoder.py::test_text_encoder_sharded fails with 
```
RuntimeError: Value out of range (expected to be in range of [-256, 255], but got -4095)
```
### What's changed

1. A new wrapper class Gemma2TextEncoderWrapper was added in model_utils.py, consistent with the existing transformer/VAE wrappers.
2. Within the wrapper's forward, use_cache=False was pinned and return_dict=False was also pinned in the wrapper so that a plain tensor (last_hidden_state) is returned for graph capture.
3. In loader.py, Gemma2TextEncoderWrapper was imported alongside the other wrappers.
4. The load_model method was updated so the text encoder is now returned wrapped: Gemma2TextEncoderWrapper(load_text_encoder(...)).
5. The load_shard_spec method was updated so the underlying model is unwrapped via model.encoder when building shard specs (mirroring how model.transformer is used for the TRANSFORMER variant).
6. Sliding window attention support will be added later for this model. 
7. Post this change, model passes e2e.

### Logs

[lumina_text_encoder_before_fix.log](https://github.com/user-attachments/files/28823425/lumina_text_encoder_before_fix.log)
[lumina_text_encoder_after_fix.log](https://github.com/user-attachments/files/28823423/lumina_text_encoder_after_fix.log)

### DIFF
--- a/lumina_image/pytorch/loader.py
+++ b/lumina_image/pytorch/loader.py
@@ -29,6 +29,7 @@ from .src.model_utils import (
     LUMINA_REPO_ID,
     MESH_NAMES,
     MESH_SHAPES,
+    Gemma2TextEncoderWrapper,
     Lumina2TransformerWrapper,
     VAEDecoderWrapper,
     load_text_encoder,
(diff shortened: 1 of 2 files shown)


infinity/pytorch/src/model.py
vkovinicTT, line 5: Why do we need whole model implementation?

meenakshiramanathan1: Removed, change is specific to lumina model.

vkovinicTT (approved):

PR #728, opened 2026-06-11
[GLM Image] Fix prior_token_drop dtype and avoid boolean-mask graph break in transformer test

meenakshiramanathan1:
### Ticket

- [#4804](https://github.com/tenstorrent/tt-xla/issues/4804)

### Problem description

tests/torch/models/glm_image/test_transformer.py::test_transformer_sharded fails with 
```
IndexError: index 4023 is out of bounds for dimension 0 with size 1
```
### What's changed

- A failure was hit in the GLM-Image transformer test , on the statement `prior_embedding[prior_token_drop] *= 0.0`.

- Initially,` IndexError: index 8692 is out of bounds for dimension 0 with size 1` was raised, because prior_token_drop was being constructed as a long tensor holding large vocabulary values, so it was interpreted as integer indices into dim-0 (size 1) rather than as a boolean mask.

- After the dtype was corrected, a second failure was surfaced during TT-device compilation: an AssertionError (`assert last_output_node is not None`) was raised by torch_xla's FX partitioner. This was caused by the boolean-mask in-place assignment being lowered to a data-dependent nonzero/gather/scatter, which produced a dynamic shape, forced a Dynamo graph break, and left a partition that could not be fused.

- The prior_token_drop input was changed from a long tensor of random vocabulary values to an all-False boolean mask, so that the real pipeline's conditional pass (prior_token_drop_cond) is matched and a valid boolean mask is provided.

- GlmImageTransformer2DModel.forward was patched so the boolean-mask in-place assignment prior_embedding[prior_token_drop] *= 0.0 is replaced by the mathematically-equivalent static-shape form prior_embedding = prior_embedding * (~prior_token_drop)[..., None].

- The data-dependent nonzero/gather/scatter was thereby eliminated, so the Dynamo graph break and the partitioner assertion were both avoided.

- Numerical correctness was preserved, since the same patched operation is executed on both the CPU golden path and the TT path, leaving PCC unaffected.

- Model now fails in mlir with `error: 'sdy.collective_permute' op requires the same type for all operands and results`

### Logs
[glm_transformer_before_fix.log](https://github.com/user-attachments/files/28824143/glm_transformer_before_fix.log)
[glm_transformer_after_fix.log](https://github.com/user-attachments/files/28824140/glm_transformer_after_fix.log)

### DIFF
--- a/glm_image/pytorch/loader.py
+++ b/glm_image/pytorch/loader.py
@@ -154,7 +154,7 @@ class ModelLoader(ForgeModel):
                                    images_per_sample (1,) int64, cache_position (394,) int64,
                                    logits_to_keep scalar int64]
         TRANSFORMER             → [hidden_states (1,16,128,144), encoder_hidden_states (1,376,1472),
-                                   prior_token_id (1,4608) int64, prior_token_drop (1,4608) int64,
+                                   prior_token_id (1,4608) int64, prior_token_drop (1,4608) bool,
                                    timestep ([999.]), target_size ([[1024.,1152.]] bf16),
                                    crop_coords ([[0.,0.]] bf16)]
         VAE                     → [z (1,16,128,144) dtype]
--- a/glm_image/pytorch/src/model_utils.py
+++ b/glm_image/pytorch/src/model_utils.py
@@ -15,6 +15,133 @@ Components:
 
 import torch
 
+
+# ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


vkovinicTT:
Have you tested the unpatched version vs patched version (both on CPU) to verify that it's correct? I mean, what happens if mask isn't all zeros or all ones?

PR #732, opened 2026-06-15
Fix FLUX.2 text encoder OOM

ashokkumarkannan1:
### Ticket
- https://github.com/tenstorrent/tt-xla/issues/5039

### Problem description
- The sharded FLUX.2-dev Mistral3 text encoder component test (`tests/torch/models/flux2/test_text_encoder.py::test_text_encoder_sharded`) fails on latest main with a DRAM **Out of Memory** on an 8-chip LLMBox:
  ```
  TT_FATAL: Out of Memory: Not enough space to allocate 335544320 B (320 MB) DRAM buffer across 12 banks
  (allocated: 1039157248 B, free: 31615936 B) — bank_manager.cpp:462
  ```
  It dies at execution (`prepareInputTensor` → `MeshBuffer::create`) because per-bank DRAM is already ~full from the encoder weights.
- Root cause is two-fold:
  1. **Shard specs never reached the decoder stack.** `shard_text_encoder_specs` walked `encoder` → `encoder.model`, landing on `Mistral3Model` — which does **not** expose `.layers` (its decoder lives one level deeper at `encoder.model.language_model`). The function therefore returned having sharded nothing in the layer stack, so the ~47 GB bf16 decoder was effectively **replicated** on every device instead of partitioned → OOM.
  2. **Mesh under-utilized the board.** The component tests run a single sample (`batch == 1`) and the shard specs partition weights only along the `model` axis. With `MESH_SHAPES[8] = (2, 4)` the `batch=2` axis shards nothing — it just replicates the (already only 4-way) sharded weights, capping effective model-parallelism at 4-way (~14 GB/device) and OOM-ing the large encoder.

### What's changed
Loader-only change in `flux2/pytorch/src/model_utils.py`:
- **Correct Mistral3 module navigation.** New `_resolve_text_transformer()` descends through the wrapper modules (`language_model` / `model` / `text_model`) until it reaches the module that actually owns `.layers`/`.embed_tokens` (`Mistral3ForConditionalGeneration → model.language_model`). `shard_text_encoder_specs` now applies its column/row-parallel specs to every decoder layer instead of silently no-op'ing.
- **Put every device on the `model` axis** so weights shard across all of them:
  ```
  MESH_SHAPES = {32: (1, 32), 8: (1, 8), 4: (1, 4), 2: (1, 2), 1: (1, 1)}
  MESH_NAMES  = ("batch", "model")
  ```
- **Impact:** the decoder weights are now genuinely sharded N-ways across all chips, so the ~47 GB bf16 encoder fits and the test runs end-to-end and passes its PCC check. Validated on an 8-chip LLMBox (mesh `(1, 8)`): the test goes from DRAM OOM → **PASSED** (~276 s).

> Note: this PR is scoped to the **text encoder**. `test_transformer_sharded` (32B) is still DRAM-bound even with 8-way sharding (OOMs on a mid-graph `TilizeWithValPadding` activation) and needs a separate memory-reduction follow-up — out of scope here.

### Checklist
- [x] Existing test provides coverage: `tests/torch/models/flux2/test_text_encoder.py::test_text_encoder_sharded` (nightly / tensor_parallel / model_test). It was xfail (PCC gap, tracking TBD); with this fix it passes and the xfail can be removed in tt-xla.

### Logs
- Before fix: DRAM OOM at input prep (320 MB buffer; `bank_manager.cpp:462`) —[test_text_encoder_sharded.log](https://github.com/user-attachments/files/28943341/test_text_encoder_sharded.log)
- After fix: `1 passed in 275.58s` on mesh `(1, 8)` — [test_text_encoder_sharded_FIX.log](https://github.com/user-attachments/files/28943336/test_text_encoder_sharded_FIX.log)


kamalrajkannan78 (approved):

flux2/pytorch/src/model_utils.py
```diff
@@ -251,14 +259,34 @@ def _add_shard_spec(specs: dict, param: torch.Tensor | None, spec: tuple) -> Non
...
+        if hasattr(module, "layers"):
+            return module
+        for attr in ("language_model", "model", "text_model"):
+            inner = getattr(module, attr, None)
+            if inner is not None and inner is not module:
+                module = inner
+                break
+        else:
+            break
+    return module
```
vkovinicTT: why magic number 4 here?

ashokkumarkannan1: @vkovinicTT Thanks for the pointing it out — the `4` wasn't derived from the model architecture, it was just an arbitrary upper bound on the descent so the loop couldn't spin forever (the real Mistral3 nesting `encoder.model.language_model` is only 2 levels). 
- There's no architectural quantity to derive it from, so rather than swap one constant for another, I removed the magic number entirely: the descent now runs `while not hasattr(module, "layers")` and stops naturally either when it reaches the decoder stack or when none of the known wrapper attrs lead anywhere new, with a `visited` id() set guarding against self-referential modules.

vkovinicTT (approved):

PR #736, opened 2026-06-15
 [LTX-2] Add pipeline-component loader

saiarthiraguram:
## Github issue
https://github.com/tenstorrent/tt-xla/issues/4541

## Problem Description

Adds `ltx2/pytorch/loader.py` covering all 6 components of the Lightricks/LTX-2
diffusion pipeline (`diffusers.pipelines.ltx2`, diffusers 0.38.0) as `ModelVariant`
entries:

| Variant | Class | Params |
|---|---|---|
| vae | AutoencoderKLLTX2Video | 1.22B |
| audio_vae | AutoencoderKLLTX2Audio | 0.053B |
| vocoder | LTX2Vocoder | 0.056B |
| connectors | LTX2TextConnectors | 1.43B |
| transformer | LTX2VideoTransformer3DModel | ~19B |
| text_encoder | Gemma3ForConditionalGeneration | ~12B |

Each variant exposes:
- `load_model()` returning the component wrapped to a clean tensors-only forward.
- `load_inputs()` building matched synthetic input tensors (e.g. connectors take
  per-layer Gemma3 hidden states `(B, seq, 3840, 49)`, 49 = 48 layers + embedding).
- `get_mesh_config` + Megatron-style `load_shard_spec` for the tensor-parallel
  `transformer` and `text_encoder`.

## Testing

- `python -m py_compile` clean; loader imports and `load_model`/`load_inputs`
  exercised for all 6 variants during bringup.
- Downstream: 3/6 components pass on TT hardware via the tt-xla tests; 3/6 xfail
  on documented compiler blockers (tenstorrent/tt-xla#5196, tenstorrent/tt-xla#5197).

### Logs
[run_audio_vae_singledev.log](https://github.com/user-attachments/files/28952674/run_audio_vae_singledev.log)
[run_connectors_singledev.log](https://github.com/user-attachments/files/28952675/run_connectors_singledev.log)
[run_text_encoder_tp.log](https://github.com/user-attachments/files/28952676/run_text_encoder_tp.log)
[run_vae_singledev.log](https://github.com/user-attachments/files/28952677/run_vae_singledev.log)
[run_vocoder_singledev.log](https://github.com/user-attachments/files/28952678/run_vocoder_singledev.log)


ltx2/pytorch/loader.py
```diff
@@ -0,0 +1,459 @@
...
+                    shard_specs[attn.to_k.weight] = ("model", None)
+                    shard_specs[attn.to_v.weight] = ("model", None)
+                    shard_specs[attn.to_out[0].weight] = (None, "model")
+                ff = getattr(block, "ff", None)
+                if ff is not None and hasattr(ff, "net"):
+                    # net[0] is GEGLU/proj (column), net[-1] is the output (row)
+                    proj = getattr(ff.net[0], "proj", None)
+                    if proj is not None:
+                        shard_specs[proj.weight] = ("model", None)
+                    shard_specs[ff.net[-1].weight] = (None, "model")
```
vkovinicTT: Please move sharding of each component into seperate function and then `load_shard_spec` will just call these functions based on the component.

vkovinicTT (changes requested):

ltx2/pytorch/loader.py
saiarthiraguram: Added changes to recent commit
### Logs
[run_audio_vae_singledev.log](https://github.com/user-attachments/files/29047738/run_audio_vae_singledev.log)
[run_text_encoder_tp.log](https://github.com/user-attachments/files/29047741/run_text_encoder_tp.log)
[run_vae_singledev.log](https://github.com/user-attachments/files/29047742/run_vae_singledev.log)
[run_vocoder_singledev.log](https://github.com/user-attachments/files/29047743/run_vocoder_singledev.log)

vkovinicTT:
I have a few questions:
- Regarding VAE did you run encoder or decoder (can we run both if possible)?
- Can we please take a deeper look into what's the issue with text encoder exactly, it might be something minor which is easy to fix?

saiarthiraguram:
Thanks @vkovinicTT — both good questions, and the second one turned out to be
right in a way I did not expect. Answers below, everything re-run on hardware
today (n300-llmbox, PCC-gated at the default 0.99).

## 1. VAE — encoder *and* decoder, both directions

The `PASS` in the original table was the **round trip**, not one half:
`_VaeWrapper` calls `AutoencoderKLLTX2Video.forward()`, which is
encode → posterior `.mode()` → decode in a single graph. That is what
`LTX2Pipeline` itself calls, so both halves were already compiled and executed —
but as one graph, which means a failure in either half could not have been
attributed to it.

So I split them. `VaeEncoder` / `VaeDecoder` (and the same for the audio VAE) are
already on this PR as of `63b6bbee32`, and I have now run all six on device:

| variant | what it runs | params | result | time |
|---|---|---|---|---|
| `Vae` | encode → mode → decode | 1.22B | **PASS** | 480 s |
| `VaeEncoder` | encode only | 0.694B | **PASS** | 74 s |
| `VaeDecoder` | decode only | 0.528B | **PASS** | 76 s |
| `AudioVae` | encode → mode → decode | 0.053B | **PASS** | 240 s |
| `AudioVaeEncoder` | encode only | — | **PASS** | 70 s |
| `AudioVaeDecoder` | decode only | — | **PASS** | 72 s |

Two details worth noting on the split wrappers:

- The encoder returns `latent_dist.parameters` — the raw Gaussian moments
  (mean ++ logvar, `(1, 256, 1, 2, 2)`), not `.mode()` — so the comparison covers
  the entire encoder output rather than half of it.
- The decoder takes latents directly and passes `temb=None`, which is correct for
  this checkpoint: `timestep_conditioning=False`.

The tt-xla side (#5203) now has these as parametrized cases
(`test_vae[round_trip|encoder|decoder]`, same for `test_audio_vae`), so all six
are covered in CI rather than only the round trip.

## 2. Text encoder — you were right, and it was already fixed upstream

It was minor, and it was not ours. Two things were wrong with how the original
xfail described it.

**It is not a sharding problem.** It reproduces unsharded, and at 2 decoder
layers instead of 48 — so the shard spec was never implicated.

**`Error code: 13` was a symptom, not the error.** The real message, from the
compiler:

```
'ttir.rms_norm' op normalized_shape dimensions must match trailing dimensions of input tensor
-> Failed to run TTIRToTTNNCommon pipeline
-> ValueError: Error code: 13
```

The mechanism: the wrapper stacks the per-layer hidden states, so a reshape that
appends a trailing unit dim (`1x128x3840` → `1x128x3840x1`) sits directly on the
final norm. `TTIRCommuteReshapeThroughRMSNorm::isCommuteUpwardsViable` gated on
`utils::preservesDim(reshapeUser, -1)`, which at that time treated a dim as
preserved when it merely moved position past size-1 dims — so appending a
trailing `1` counted as preserving the last dim. Commuting the reshape above the
norm then left `ttir.rms_norm` with input `1x128x3840x1` and
`normalized_shape=[3840]` carried over verbatim, i.e. the normalized dim was no
longer trailing, and the op's own verifier rejected it.

That is exactly tt-mlir
[#8730](https://github.com/tenstorrent/tt-mlir/pull/8730) /
[issue #7150](https://github.com/tenstorrent/tt-mlir/issues/7150) — *"[EIO]
Reject rms_norm reshape commute that breaks trailing dim"* — whose description is
this failure verbatim. It renamed the loose check to `preservesDimUpToOnes` and
reclaimed `preservesDim` for a stricter one that also requires the output dim at
the same offset from the trailing edge to match. **Merged 2026-06-09.**

Our `tt-mlir` pin during the bringup was `c5f398432` (**May 25**) — two weeks
older than the fix. Re-run on the current pin, `e2b21a7219` (Aug 24), which
contains it:

| run | result | verifier hits |
|---|---|---|
| 2-layer repro, unsharded | **PASS** (215 s) | 0 |
| 2-layer repro, full Megatron shard map | **PASS** (102 s) | 0 |
| **full Gemma3-12B, 8-way TP** | **PASS** (112 s, mesh `(1, 8)`) | 0 |

So the text encoder is green, not xfail. I have removed the marker on the tt-xla
side and tt-xla#5197 can be closed as fixed by tt-mlir#8730.

I cannot prove *only* #8730 was responsible — three months of tt-mlir separate
the two pins — but the error string, the shape signature and the pass are all an
exact match for it, and the failure is gone on the first pin that contains it.

## Updated component status

While re-running these I re-checked the other two xfails as well, since both of
their blockers have since been closed upstream:

| Component | Mode | On this PR | Today |
|---|---|---|---|
| `vae` — round trip / encoder / decoder | single_device | PASS (round trip only) | **PASS** ×3 |
| `audio_vae` — round trip / encoder / decoder | single_device | PASS (round trip only) | **PASS** ×3 |
| `vocoder` | single_device | PASS | **PASS** (220 s) |
| `connectors` | single_device | xfail #5196 | **PASS** (219 s) |
| `text_encoder` | tensor_parallel (1, 8) | xfail #5197 | **PASS** (112 s) |
| `transformer` | tensor_parallel (1, 8) | xfail #5196 | still failing, new error |

`connectors` is the loader's XLA-safe rope replacement finally confirmed on
hardware — it had been written but never verified, so this is the first run that
demonstrates it. **9 of the loader's 10 variants now pass** (the VAE split added
four); #736's table covered 6, of which 3 passed.

The `transformer` is the one remaining gap, and it has moved: it now clears the
`ManualComputationOp` fragmentation that used to block it (tt-mlir#9014, closed —
fixed by #8601), compiles all the way through, and fails in *execution* instead:

```
TT_FATAL: Can't get a single buffer from host storage distributed over mesh
shape MeshShape([1, 8])   (tensor_apis.cpp:631, buffers.size() == 1)
  ttnn::distributed::distribute_tensor <- DistributeTensorOp <- ProgramExecutor::execute
```

The runtime is trying to distribute a host tensor that is already distributed
over the mesh. Strictly later than anything previously recorded, and not
loader-side — I will file it separately rather than leave it under #5196.

## One process note

I have opened this as a fresh PR — tenstorrent/tt-forge-models#923 — and am
closing this one. The loader content is the same, but it is now rebased on
current `main` and every row in the table above is a run from today, whereas
this PR has three months of drift and a result table that is wrong in three
places. Easier to review the new one than to reconcile this thread. Apologies
for the delay in getting back to you.

saiarthiraguram:
Superseded by #923 — same loader, rebased on current main, all results re-run today. Review thread continues there.

PR #737, opened 2026-06-15
Add LongCat-Image per-component model loader

saiarthiraguram:
Brings up meituan-longcat/LongCat-Image (a LongCatImagePipeline, ~14 B
aggregate bilingual text-to-image MMDiT) on a single n150, per-component.

https://github.com/tenstorrent/tt-xla/issues/4698

Adds a per-component loader for [`meituan-longcat/LongCat-Image`](https://huggingface.co/meituan-longcat/LongCat-Image), a bilingual text-to-image MMDiT pipeline (~14 B aggregate).

### Problem Description
LongCat-Image is a `LongCatImagePipeline` with three parameterized components. Rather than loading the full pipeline, each `ModelVariant` scaffolds **one component** as an independent, tensors-only `torch.nn.Module` so the runner can compile and PCC-compare it in isolation. Each component is fetched directly via `from_pretrained(..., subfolder=...)` — the full pipeline is never instantiated.

| Variant | Component | Forward |
|---|---|---|
| `LongCat_Image_TextEncoder` | Qwen2.5-VL 7B (`Qwen2_5_VLForConditionalGeneration`) | `(input_ids, attention_mask) → last hidden state [B, 553, 3584]` |
| `LongCat_Image_Transformer` | `LongCatImageTransformer2DModel` (Flux-style, ~6 B; 10 dual + 20 single blocks) | `(hidden_states, timestep, encoder_hidden_states) → [B, 256, 64]` |
| `LongCat_Image_Vae` | `AutoencoderKL` (16-ch, 8× spatial compression) | `decode(latent) → image` |

### Notes
- **Position ids are reconstructed, not fed.** The transformer's structural `img_ids`/`txt_ids` are rebuilt analytically via a port of `LongCatImagePipeline.prepare_pos_ids` (verified bit-exact against the captured tensors) and pinned as non-persistent buffers, so `forward` takes only data tensors. `guidance` is `None` (guidance embeds disabled in this checkpoint).
- **Input shapes are real captured I/O** from a single CPU bf16 forward at 256×256 / 2 denoise steps with prompt-rewrite disabled. `load_inputs()` builds synthetic tensors at those exact shapes.
- **Requires `diffusers >= 0.36`** (LongCat classes first released in 0.36.0) — see `longcat_image/requirements.txt`.
- The loader also records the tensor-parallel shard plan (`SHARD_SPECS` / `TT_VISIBLE_DEVICES`) for the multi-chip follow-up: the 6 B transformer and 7.7 B text encoder exceed a single n150 (7 B/chip); the VAE fits.

### Logs

[hw_text_encoder.log](https://github.com/user-attachments/files/28827242/hw_text_encoder.log)
[hw_transformer.log](https://github.com/user-attachments/files/28827243/hw_transformer.log)
[hw_vae.log](https://github.com/user-attachments/files/28827244/hw_vae.log)Logs


longcat_image/pytorch/loader.py
```diff
@@ -0,0 +1,315 @@
...
+  scheduler     -> FlowMatchEulerDiscreteScheduler (no parameters, skipped)
+
+Each variant scaffolds one component as an independent torch.nn.Module the
+runner can compile + PCC-compare in isolation. The full pipeline is never
+loaded -- each component is fetched directly via from_pretrained(..., subfolder=...).
+
+Shapes below are the REAL captured I/O of a single CPU forward at 256x256 /
+2 denoise steps with prompt-rewrite disabled (see _COMPONENT_IO_SPEC and
+.claude/bringup/meituan-longcat__LongCat-Image/io_spec_raw.json). The
+transformer's non-tensor / structural position-id args (img_ids / txt_ids /
```
vkovinicTT: Please remove this, it's local claude file and doesn't belong here.

vkovinicTT (changes requested):

longcat_image/pytorch/loader.py
saiarthiraguram: Removed it in recent commit @vkovinicTT

vkovinicTT (approved):

PR #746, opened 2026-06-18
Add loader for Seamless_m4t_v2 Pytorch Model

meenakshiramanathan1:
### Ticket

- [#5218](https://github.com/tenstorrent/tt-xla/issues/5218)

### Problem description

- Bringup Seamless_m4t_v2 pytorch model

### What's changed

- Added the loader script for mentioned variant

### Checklist
- [x] Verified the changes through local testing

### Log

[seamless_v2.log](https://github.com/user-attachments/files/29366602/seamless_v2.log)

### DIFF
--- a/seamless_m4t_v2/pytorch/requirements.txt
+++ b/seamless_m4t_v2/pytorch/requirements.txt
@@ -1,0 +1,4 @@
+--extra-index-url https://download.pytorch.org/whl/cpu
+torchaudio==2.10.0
+torchcodec==0.10.*
+datasets
--- a/seamless_m4t_v2/pytorch/__init__.py
+++ b/seamless_m4t_v2/pytorch/__init__.py
@@ -1,0 +1,7 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""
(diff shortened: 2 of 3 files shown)


seamless_m4t_v2/pytorch/requirements.txt
```diff
@@ -0,0 +1,4 @@
+--extra-index-url https://download.pytorch.org/whl/cpu
```
nvukobratTT, line 1: Why is this one requred?

nvukobratTT (approved):
Small comment to address

seamless_m4t_v2/pytorch/requirements.txt
meenakshiramanathan1: tt-xla's venv is CPU-only (torch==2.10.0+cpu, no libtorch_cuda.so in torch/lib, built from download.pytorch.org/whl/cpu). Per-model requirements are pip-installed against PyPI by default, and PyPI's torchaudio==2.10.0 is the CUDA build — its extension links libtorch_cuda.so, which doesn't exist here, so import torchaudio dies with `libtorch_cuda.so: cannot open shared object file` . --extra-index-url https://download.pytorch.org/whl/cpu pulls the +cpu torchaudio wheel that links libtorch_cpu.so, matching the venv's torch.

PR #747, opened 2026-06-18
Qwen2.5-Math-72B-Instruct variant Bringup

sonalibaskaran2499:
### Ticket
Link to Github Issue

### Problem description
Add the `Qwen/Qwen2.5-Math-72B-Instruct` variant to the Qwen2.5 causal-LM loader.
At bf16 the 72B weights (~145 GB) don't fit across the target 4-device mesh (~36 GB/device vs ~34 GB available), so the model OOMs.

### What's changed
- Added the `Math_72B_Instruct` variant (`Qwen/Qwen2.5-Math-72B-Instruct`) to  `ModelVariant` and `_VARIANTS`.
- Added a `mixed_precision_configs/Qwen2.5-Math-72B-Instruct.json` (`{"default": "bfp_bf8"}`) so the weights are converted to **bfp8**, halving weight memory to ~18 GB/device and letting the model fit and run.
- Sharded the token embedding (`embed_tokens.weight`) in `load_shard_spec` so the large vocab table is no longer replicated on every device.

### Checklist

### DIFF
--- a/qwen_2_5/causal_lm/pytorch/loader.py
+++ b/qwen_2_5/causal_lm/pytorch/loader.py
@@ -42,6 +42,7 @@ class ModelVariant(StrEnum):
     QWEN_2_5_72B_INSTRUCT = "72B_Instruct"
     QWEN_2_5_72B = "72B"
     QWEN_2_5_MATH_7B = "Math_7B"
+    QWEN_2_5_MATH_72B_INSTRUCT = "Math_72B_Instruct"
 
 
 class ModelLoader(ForgeModel):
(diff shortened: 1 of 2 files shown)


qwen_2_5/causal_lm/pytorch/loader.py
```diff
@@ -277,6 +282,7 @@ def get_mesh_config(self, num_devices: int):
     def load_shard_spec(self, model):
         """Default shard spec on a ("batch", "model") mesh."""
         shard_specs = {}
+        shard_specs[model.model.embed_tokens.weight] = ("model", "batch")
```
kmabeeTT: This safe for the other variants too? Just curious if this new variant passes and existing variants still pass?

kmabeeTT (approved):
One question otherwise looks good

qwen_2_5/causal_lm/pytorch/loader.py
sonalibaskaran2499: @kmabeeTT  The embedding-sharding change was applied to all variants in the first version. I've now gated it to only the new Qwen2.5-Math-72B variant, so the existing variants shard spec is completely unchanged and they pass exactly as before.

PR #768, opened 2026-06-21
Bringup Orion-14B-Chat model

sonalibaskaran2499:
### Ticket
Link to Github Issue

### Problem description
Bringup Orion-14B-Chat model.

### What's changed
Added Orion-14B-Chat tensor-parallel loader.

### Note on the transformers pin
Orion uses `trust_remote_code`; its modeling file relies on transformers APIs removed in the repo default 5.5.1 (`is_flash_attn_2_available`, `_tied_weights_keys`). 4.57.1 is the last release it runs against. Proper fix is upstream; pin is installed only for this model and rolled back after.

### Logs
[orion.log](https://github.com/user-attachments/files/29173378/orion.log)

### Checklist

### DIFF
--- a/orion/pytorch/requirements.txt
+++ b/orion/pytorch/requirements.txt
@@ -1,0 +1,1 @@
+transformers==4.57.1
--- a/orion/pytorch/requirements.txt
+++ b/orion/pytorch/requirements.txt
@@ -1,1 +1,4 @@
+# Orion's trust_remote_code modeling file uses transformers APIs removed in
+# 5.5.1 (repo default), e.g. is_flash_attn_2_available and _tied_weights_keys.
+# 4.57.1 is the last release it imports/runs cleanly against.
 transformers==4.57.1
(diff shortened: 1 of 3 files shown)


orion/pytorch/requirements.txt
```diff
@@ -0,0 +1 @@
+transformers==4.57.1
```
kmabeeTT, line 4: Why was this needed? I noticed this in a few PR's lately, is there some breakage with newer transformers? We should document what it is at least in the PR Description or comments for posterity, and consider what the fixes could be.

kmabeeTT (commented):
Question about the transformers downgrade pin, I thought it might be a one-off from a model @devisettymahidhar608 worked on, but now it's popping up more in new models..

orion/pytorch/requirements.txt
sonalibaskaran2499: Added a comment in requirements.txt and also updated the PR description explaining it . Same root cause across these remote-code models (Orion, InternLM2, etc.): their trust_remote_code modeling files were written for transformers 4.x and break on the repo default 5.5.1 (removed is_flash_attn_2_available, _tied_weights_keys→all_tied_weights_keys rename). 4.57.1 is the last release Orion's remote code runs against. The proper fix is upstream is the model author updating the remote code for 5.5.1. So the per-model pin (installed only for this model, then rolled back) is the pragmatic path, mirroring what the JAX models already do.

kmabeeTT (approved):

PR #775, opened 2026-06-23
Fix sdy.collective_permute type mismatch in GLM-image transformer sharding

meenakshiramanathan1:
### Problem description

`tests/torch/models/glm_image/test_transformer.py::test_transformer_sharded` fails with:

`loc("add.3248"): error: 'sdy.collective_permute' op requires the same type for all operands and results`

### What's changed

- All changes are in shard_transformer_specs in third_party/tt_forge_models/glm_image/pytorch/src/model_utils.py.

- Single-axis "model" Megatron. Every collective is now kept on a single axis (all_gather / all_reduce only, never collective_permute):

- Attention + FFN → plain single-axis Megatron on "model". Column-parallel (to_q/to_k/to_v, ff.net[0]) changed from ("model","batch") to ("model", None); row-parallel (to_out, ff.net[2]) changed from ("batch","model") to (None, "model"). The residual stream (hidden_states / encoder_hidden_states) therefore stays replicated: column layers consume it gather-free, and row layers all_reduce back to replicated before the residual add.

- AdaLN modulation linears → replicated. norm1.linear / norm1_context.linear (per block) and norm_out.linear emit N*dim outputs that are .chunk()'d (12× for norm1, shift/scale for norm_out). Sharding that output on "model" (size 4) makes chunk boundaries straddle shard boundaries, forcing the per-chunk reshard that produced the flood of slice collective_permutes. All changed to (None, None) weight / (None,) bias to keep them replicated.

- Entry projectors / embeddings feeding the residual stream → replicated. image_projector.proj → (None, None); prior_token_embedding → (None, None); proj_out → (None, None); time/size conditioning embedders (timestep_embedder, condition_embedder linears) → (None, None). Their output must match the replicated residual state.

- glyph / prior projectors → column→row Megatron on "model". These FeedForward modules now do their own ("model", None) up / (None, "model") down internally so their output lands replicated. Also refactored glyph_projector and prior_projector handling into a single shared loop.

- With these changes the sharded GLM-Image transformer test now compiles and runs end-to-end on 8 devices.

### Logs
[glm_transformer_before_fix_june23.log](https://github.com/user-attachments/files/29233115/glm_transformer_before_fix_june23.log)
[glm_transformer_after_fix_june23.log](https://github.com/user-attachments/files/29233130/glm_transformer_after_fix_june23.log)

### DIFF
--- a/glm_image/pytorch/src/model_utils.py
+++ b/glm_image/pytorch/src/model_utils.py
@@ -668,46 +668,48 @@ def shard_vae_specs(vae) -> dict:
 def shard_transformer_specs(transformer) -> dict:
     """Shard specs for GlmImageTransformer2DModel.
 
-    Column-parallel (Q, K, V, FFN up, modulation linear): ("model", "batch")
-    Row-parallel   (O, FFN down, proj_out):               ("batch", "model")
+    Single-axis "model" Megatron (mesh is ("batch", "model"), but data
+    batch == 1). Using "batch" as a second weight-sharding axis would force a
+    model<->batch axis-swap reshard that lowers to collective_permute.
+
+    Attention + FFN: column-parallel (to_q/k/v, ff up) -> ("model", None);
+    row-parallel (to_out, ff down) -> (None, "model"). The residual stream stays
+    REPLICATED (row layers all_reduce back before the residual add). AdaLayerNorm
+    modulation linears also stay REPLICATED: their outputs are `.chunk()`'d, and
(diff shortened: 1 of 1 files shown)


nvukobratTT (approved):

meenakshiramanathan1:
This PR will be made ready for review once this tt-mlir [commit](https://github.com/tenstorrent/tt-mlir/commit/688a8a811a27aa36f428133dd769b91e0a65a540) is tracked in tt-xla.

PR #776, opened 2026-06-23
Add VibeVoice-1.5B PyTorch loader (port mode)

saiarthiraguram:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/5217)

### Problem description

[`microsoft/VibeVoice-1.5B`](https://huggingface.co/microsoft/VibeVoice-1.5B) — a long-form, generation-based text-to-speech (multimodal TTS) model — has no tt-forge-models loader. It cannot be brought up as-is: the HuggingFace repo ships **weights only** (no remote code), and the architecture (`model_type: vibevoice`) is **not in transformers**, so there is no off-the-shelf way to instantiate the model. The full model is also ~2.7B params (the "1.5B" refers only to the Qwen2.5 LM backbone; the rest is acoustic/semantic VAE tokenizers + diffusion head + connectors), and its dataclass forward output contains a non-tensor int leaf that breaks the inference harness's pytree comparator.

### What's changed

Adds a PyTorch loader for VibeVoice-1.5B in **port mode**: the core model code is vendored from [github.com/microsoft/VibeVoice](https://github.com/microsoft/VibeVoice) `@303b2833` (provenance + file list in `vibevoice/pytorch/src/SRC_VENDORED_FROM.txt`).

- `ModelLoader` / `ModelVariant` return the full VibeVoice model with **random weights** built from a vendored real 1.5B `config.json` (config-only — no weight download).
- `_VibeVoiceLogitsWrapper` exposes a clean **tensors-only** `forward()` returning just `logits`, so the harness's pytree comparator doesn't trip on the dataclass int leaf (`speech_token_num`) under `torch.equal`.
- `transformers>=5` compat shims, documented inline: `tie_weights(self, *args, **kwargs)` (v5 passes `recompute_mapping=`) and idempotent `AutoModel(.ForCausalLM).register(exist_ok=True)` (v5 pre-registers configs).
- Bringup forward path uses `speech_tensors=None` → reduces to a clean Qwen2.5 LM forward → logits. `speech_semantic_tensors (1,1,128)` is required (semantic connector is called unconditionally) but its result is unused.

**Impact:** VibeVoice-1.5B can now be loaded and run through the standard tt-forge-models flow. Brought up single-device on **n150 (inference)** via tt-xla, enabling the companion tt-xla test-config PR.

Layout:
```
vibevoice/
├── __init__.py
└── pytorch/
    ├── __init__.py
    ├── loader.py
    ├── config.json          # vendored real 1.5B arch (config-only; random weights)
    └── src/
        └── vibevoice/{modular,schedule}/...   # 6 core files vendored from upstream
```

Notes: vendored `dpm_solver.py` keeps its TSAIL/HuggingFace Apache-2.0 notice (SPDX tag added); other ported files carry the standard Tenstorrent Apache-2.0 header per repo convention (matches the `maptr`/`transfuser`/`bevformer` vendoring pattern). The empty `__init__.py`s under `src/vibevoice/` intentionally override upstream ones (which pull in streaming/processor + gradio/av/aiortc and double-register Auto classes).

### Checklist
- [x] New/Existing tests provide coverage for changes — single-device n150 inference passes via `test_all_models_torch[vibevoice/pytorch-1.5B-single_device-inference]` (`1 passed`, default PCC); CPU sanity produces finite logits `(1, 32, 151936)`.

### Logs
[bringup_steps.txt](https://github.com/user-attachments/files/29241006/bringup_steps.txt)
[iter_2_run.log](https://github.com/user-attachments/files/29241008/iter_2_run.log)
[model_overview.md](https://github.com/user-attachments/files/29241010/model_overview.md)


vibevoice/pytorch/loader.py
```diff
@@ -0,0 +1,223 @@
...
+        # non-tensor leaves in VibeVoiceCausalLMOutputWithPast).
+        return _VibeVoiceLogitsWrapper(model.eval()).eval()
+
+    def load_inputs(self, batch_size=1, seq_len=32, dtype_override=torch.bfloat16):
+        """Load and return sample inputs for the VibeVoice model.
+
+        The bringup forward path keeps ``speech_tensors=None`` so the model
+        behaves as a Qwen2.5 causal LM with a single (unused) semantic
+        connector call. ``speech_semantic_tensors`` is therefore required but
+        its result is not consumed downstream.
```
vkovinicTT, line 260: Why don't we run the other components of the model, but just `Qwen2.5 LM`? Or maybe I am misinterpreting something.

saiarthiraguram: @vkovinicTT You're reading it right, this bringup exercises the Qwen2.5 LM backbone (embeddings → decoder → lm_head), plus the one semantic_connector call that forward() makes unconditionally. The other components are deliberately left out, and it's a structural limitation rather than a shortcut:

1. The diffusion head + noise scheduler are generation/training-only. In VibeVoiceForConditionalGeneration.forward(), prediction_head and noise_scheduler only run inside the if speech_tensors is not None and acoustic_loss_mask.sum() > 0: diffusion-loss block. The actual audio synthesis lives in the autoregressive generate() loop with multi-step DPM-Solver diffusion sampling. That's iterative, data-dependent control flow, not a single static tensor-in/tensor-out graph, so it doesn't fit the single-device inference bringup contract (one traceable graph, raw-output PCC compare).

2. The acoustic tokenizer (VAE) needs real audio. forward_speech_features() only calls acoustic_tokenizer.encode(...) when speech_tensors is provided, i.e. real waveform input. It's a distinct sub-model (the acoustic VAE) and is better brought up as its own component, the way we decompose other pipeline models — not folded into the LM forward.

3. Passing speech_tensors would inject data-dependent scatter. The audio path does x[acoustic_input_mask] = speech_connect_features, boolean-mask index assignment driven by runtime data, which is exactly the kind of dynamic op that breaks clean StableHLO single-graph compilation.

So with speech_tensors=None we get the static, compiler-friendly slice of the model; which is also the dominant compute. speech_semantic_tensors is still required because forward() always runs semantic_connector(speech_semantic_tensors); its output just isn't scattered into the embeddings on this path (hence "required but not consumed downstream").

If we want device coverage of the acoustic VAE / diffusion head, the right follow-up is separate per-component bringups rather than widening this single forward. Happy to file those as follow-ups.

nvukobratTT (commented):
A general comment: why don't we just submodule a third-party GitHub project and load the respective components as before?

saiarthiraguram:
> A general comment: why don't we just submodule a third-party GitHub project and load the respective components as before?

We followed the third_party submodule pattern for Toyota repo @nvukobratTT , in tt-forge-models, we ported the required code.
However, I have refactored and added the repo as third_party submodule in the latest commit. Let me know if this works!

Logs:
[hw_test_20260701_180138.log](https://github.com/user-attachments/files/29562918/hw_test_20260701_180138.log)

nvukobratTT (approved):

PR #663, opened 2026-05-15
Add kimi-k2.5 (modified)

gengelageTT:
### Problem description
We want to bring up kimi-k2.6. Since kimi-k2.5 (modified) is compiling and running for 2 layers through the benchmark tester, we can start by adding this. The modifications to this model are necessary to run this model through tt-xla but all modifications have been marked with `# Modified: <reason> `.

### What's changed
Added files and loader for kimi-k2.5 including modified model files which follow the same modifications made for kimi-k2. 

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #779, opened 2026-06-23
Add kimi-k2.6 and weight loading for kimi-k2 variants

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/4351 

### Problem description
We want to bringup kimi-k2.6 starting with a benchmark with a couple layers.

### What's changed
Added loader and modified model files for kimi-k2.6 (matches kimi-k2.5).

Added meta_loading.py for all kimi-k2 variants so models meta-build first and load only the first `num_layers` of weights from their bf16 checkpoints, rather than loading random weights.

### Checklist


gengelageTT:
Kimi-k2 with real weights passed with PCC of `0.999` (`0.995` in latest nightly) here: https://github.com/tenstorrent/tt-xla/actions/runs/27844137330/job/82417983892#step:15:347 

Kimi-k2.5 and kimi-k2.6 running here and passing locally with similar pcc: https://github.com/tenstorrent/tt-xla/actions/runs/28032078854

kimi_k2/k2_6/pytorch/original_modeling_deepseek.py
```diff
@@ -0,0 +1,1869 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```
kmabeeTT, line 1: In previous kimi PR (https://github.com/tenstorrent/tt-forge-models/pull/582) we didn't check in the original code, but I guess it doesn't hurt to have this.  Wasn't sure about TT SDPX header on top of apache license, but looking around this seems to be the convention.

kmabeeTT (approved):
Thanks - seems good. Thanks for including passing links.

kimi_k2/k2_6/pytorch/original_modeling_deepseek.py
gengelageTT: I was asked to add original versions of the code after merging that previous kimi PR, so I added the original code for kimi-k2 and k2.5 in https://github.com/tenstorrent/tt-forge-models/pull/663 and deepseekv3.1 and v3.2-exp also have their original model files. 

If I don't add the TT SDPX header then pre-commit will fail - which is likely why other files do it as well. Not sure if it's correct or not though.

PR #781, opened 2026-06-24
[lumina_image] Use single-axis Megatron for transformer sharding

### LINKED ISSUE #5653: [Lumina Image] PCC Drop in Transformer

meenakshiramanathan1:
### Ticket
Fixes [#5653](https://github.com/tenstorrent/tt-xla/issues/5653)

### Problem description
`tests/torch/models/lumina_image/test_transformer.py::test_transformer_sharded` compiles and runs on 8 devices, but the sharded output fails with a  PCC drop of 0.093.

### What's changed

- The strategy was redesigned with the sharding-model-analysis skill: the model architecture was mapped from the config and diffusers source, the mesh was fixed at (2,4)=8 devices, and the collective for each candidate sharding pattern was looked up in the skill's CCL cheatsheet before any code was touched.

- Attention and feed-forward Megatron pairs. Switched from the 2-D alternating scheme to single-axis Megatron on "model": column-parallel ("model", None) for to_q/to_k/to_v and feed_forward.linear_1/linear_3, row-parallel (None, "model") for to_out[0] and feed_forward.linear_2. Every contraction dim now stays on one axis, so each pair emits only an all-reduce and no cross-axis reshard — removing the numerically-corrupt resharding that was collapsing PCC. GQA is preserved, as the 24 query heads and 8 kv heads are both divisible by the model-axis size. Applied generically across all 30 blocks (noise_refiner, context_refiner, layers).

- AdaLN modulation linear norm1.linear. Its output is a concatenated modulation vector chunked into scale/gate slices that multiply a replicated normalized activation. Sharding Cout on "model" made every sliced chunk inherit "model", conflicting with the replicated activation it modulates and forcing a reshard. norm1.linear is removed from the spec (left replicated) so the chunks match the activation with no collective. Applied generically across all modulation blocks.

- Boundary embedders and output projection. x_embedder, time_caption_embed (timestep_embedder, caption_embedder) and norm_out (linear_1/linear_2) had been sharded with the same alternating 2-D pattern. They sit at graph boundaries and are tiny relative to the blocks, so they are left replicated (removed from the spec); this keeps block inputs/outputs replicated and avoids any boundary reshard or final all-gather.

With these changes, the pcc jumps to 0.93.

### DIFF
--- a/glm_image/pytorch/src/pipeline.py
+++ b/glm_image/pytorch/src/pipeline.py
@@ -21,9 +21,9 @@ path) with an explicit CPU/TT device split, reusing the diffusers pipeline's own
 helper methods (``generate_prior_tokens``, ``encode_prompt``, ``prepare_latents``
 and the scheduler) so only the device split is bespoke:
 
-  - DiT transformer on Tenstorrent, tensor-parallel sharded on the
-    ``("batch", "model")`` mesh (Megatron column/row from
-    ``shard_transformer_specs``; see ``model_utils``).
+  - DiT transformer on Tenstorrent tensor-parallel sharded on the
+  ``("batch", "model")`` mesh (Megatron
+    column/row from ``shard_transformer_specs``; see ``model_utils``).
   - AR prior-token generation, T5 glyph encoding, the FlowMatchEuler scheduler
     and the VAE decode all stay on CPU.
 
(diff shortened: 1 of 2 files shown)


nvukobratTT (approved):

meenakshiramanathan1:
This PR will be made ready for review when dependent tt-mlir PR is merged

PR #784, opened 2026-06-25
Bringup XTTS-v2 (coqui/XTTS-v2) text-to-speech model

ashokkumarkannan1:
### Ticket
- https://github.com/tenstorrent/tt-xla/issues/5216 (Model Bringup: coqui/XTTS-v2)

### Background / how this PR evolved
This was **initially brought up as a single `ModelLoader` + one comparison test**: a deterministic monkey-patched `Xtts.forward` running the traceable tail (GPT latents → HiFiGAN) over a fixed/synthetic `gpt_codes` tensor, PCC-compared against native `Xtts.inference` (PCC 1.0 on CPU).

Per @nvukobratTT's review — *"We should strive to have a text-to-speech run. Just having a valid PCC for the model isn't enough as end results are not reproducible"* — it was reworked to **bring up every learned `nn.Module` as its own component and run them as a real end-to-end text→speech pipeline** (no synthetic codes).

### What's changed (current state)
- **Per-component loader** (`xtts_v2/pytorch/`): one `ModelLoader` exposing 6 single-forward variants, each fed the real tensors it receives end-to-end (derived deterministically from a real reference clip — no synthetic inputs):
  - `speaker_encoder`, `conditioning`, `gpt_prefill`, `gpt_decode`, `gpt_latents`, `hifigan_decoder`
- **Reusable e2e pipeline** (`pipeline.py`): `XTTSPipeline` chains all learned modules on TT for a real text→speech run (tokenizer, mel front-end and token sampling stay on CPU). Shared by the tt-xla e2e test.
- **`GptCachedStep`** (`src/model.py`): the static-KV-cache decode step the autoregressive loop reuses (fixed shapes → single-compile, no per-step recompile). Shared by the pipeline and the tt-xla decode-loop test.
- Kept in float32 (XTTS submodules don't cast uniformly to bf16); `isin_mps_friendly` shim for `transformers>=5`.

### Checklist
- [x] New/Existing tests provide coverage for changes (tt-xla side: component config + decode-loop test + e2e test + benchmark)

### Logs
- [hifigan_log_after_pcc_threshold.log](https://github.com/user-attachments/files/29968073/1b_hifigan_098.log), [all_components.log](https://github.com/user-attachments/files/29968070/1_components.log)


kamalrajkannan78:
Add below contents in `xtts_v2/pytorch/src/__init__.py`

```
# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
#
# SPDX-License-Identifier: Apache-2.0
```

kamalrajkannan78 (approved):

xtts_v2/pytorch/loader.py
```diff
@@ -0,0 +1,149 @@
...
+        speaker = self.model.speaker_manager.speakers[self.DEFAULT_SPEAKER]
+        gpt_cond_latent = speaker["gpt_cond_latent"].to(model_dtype)
+        speaker_embedding = speaker["speaker_embedding"].to(model_dtype)
+
+        text_tokens = torch.IntTensor(
+            self.model.tokenizer.encode(text.strip().lower(), lang=language)
+        ).unsqueeze(0)
+        text_len = torch.tensor([text_tokens.shape[-1]])
+
+        # Fixed synthetic GPT audio codes instead of running the autoregressive
```
nvukobratTT: What is the main diff between this override and the .generate function?

nvukobratTT (approved):

xtts_v2/pytorch/loader.py
ashokkumarkannan1: @nvukobratTT thanks for the review
The override isn't a reimplementation of `gpt.generate` — it deliberately **replaces it**.  Native `Xtts.inference` runs three things: 
```
(1) `gpt.generate` — the autoregressive sampling loop that produces `gpt_codes` token-by-token (incremental KV-cached GPT2 decode + `mel_head` logits + top_k/top_p/temperature sampling); 
(2) a full-sequence `gpt(..., return_latent=True)` over those codes → latents; 
(3) HiFiGAN decode → waveform.
```


- Stage (1) can't go into a tt-xla graph: it's a **dynamic-length, stochastic host loop** (loops until a stop token, samples each step), so it's neither static nor reproducible for PCC. 
- So the patched `forward` keeps only the **deterministic, traceable tail — stages (2) + (3)** exactly as inference runs them, and feeds a **fixed (seeded) random `gpt_codes` tensor** in place of the sampled one. 
- The graph only needs valid audio-token ids of the right shape, so this exercises the identical compute path (GPT2 trunk → latents → vocoder) while staying reproducible.

### Net difference vs the original generate:
a) **Skipped:** the autoregressive sampling loop + KV-cache incremental decode, and `mel_head` (the audio-token logit head used only inside generation).
b)  **Exercised (same as native):** GPT2 transformer trunk, embeddings, `final_norm`, and the HiFiGAN decoder — the heavy modules.



- Since the codes are synthetic, this is a graph/PCC bring-up of the deterministic path, not a real text→speech run.
-  If we want real audio without changing the device graph, we can run the (non-traceable) `gpt.generate` on CPU with a fixed seed and feed those real codes instead of `torch.randint` — same traced graph, meaningful output. 
- Equivalence of the override to native `Xtts.inference` is PCC 1.0 on CPU.

nvukobratTT: We should strive to have a text-to-speech run. Just having a valid PCC for the model isn't enough as end results are not reproducible

ashokkumarkannan1: Sure, will modify to work as a e2e pipeline run

acicovicTT (changes requested):
I tried running the whole pipeline on n150 at the latest tt-xla `main`, and the compilation failed with:
```
loc("custom-call.289"): error: 'ttnn.group_norm' op flattened height must be tile-aligned, got 505
2026-07-16 14:42:56.967 ( 138.336s) [        6D0D1140]      module_builder.cc:1113   ERR| Failed to run TTIRToTTNNCommon pipeline
```
This is the command I ran:
```
python -u -m third_party.tt_forge_models.xtts_v2.pytorch.pipeline \
  --output tmp/xtts_ref/xtts_output.wav
```

I managed to work around it locally, in `pipeline.py`, but this probably needs a fix in tt-mlir:

```
Workaround currently applied (to pipeline.py only)

Host-side pad of the mel time axis up to the next multiple of 32 (edge-replication) before the conditioning encoder. The downstream Perceiver resampler collapses the time axis to a fixed 32 latents, so the [1, 1024, 32] conditioning-latent output shape is unchanged. With this, the conditioning group_norm error clears and the full e2e pipeline runs on TT and produces a 24 kHz WAV.
```

ashokkumarkannan1:
> I tried running the whole pipeline on n150 at the latest tt-xla `main`, and the compilation failed with:
> 
> ```
> loc("custom-call.289"): error: 'ttnn.group_norm' op flattened height must be tile-aligned, got 505
> 2026-07-16 14:42:56.967 ( 138.336s) [        6D0D1140]      module_builder.cc:1113   ERR| Failed to run TTIRToTTNNCommon pipeline
> ```
> 
> This is the command I ran:
> 
> ```
> python -u -m third_party.tt_forge_models.xtts_v2.pytorch.pipeline \
>   --output tmp/xtts_ref/xtts_output.wav
> ```
> 
> I managed to work around it locally, in `pipeline.py`, but this probably needs a fix in tt-mlir:
> 
> ```
> Workaround currently applied (to pipeline.py only)
> 
> Host-side pad of the mel time axis up to the next multiple of 32 (edge-replication) before the conditioning encoder. The downstream Perceiver resampler collapses the time axis to a fixed 32 latents, so the [1, 1024, 32] conditioning-latent output shape is unchanged. With this, the conditioning group_norm error clears and the full e2e pipeline runs on TT and produces a 24 kHz WAV.
> ```

pipeline [test](https://github.com/tenstorrent/tt-xla/pull/5376/changes#diff-c61a99c928ed0423b3fc2e40f928f5535eb894e8b7d2f8ac0ab0636099188086R7) with real inputs has non-tile-aligned inputs and this mlir fix will solve the issue - https://github.com/tenstorrent/tt-mlir/pull/8935

acicovicTT (approved):
@ashokkumarkannan1 Understood, pls just make sure nothing new shows up in nightly as failing. If the component needs a tt-mlir fix, mark it as xfailed and leave a reference to the tracing issue in the test config file.

ashokkumarkannan1:
> @ashokkumarkannan1 Understood, pls just make sure nothing new shows up in nightly as failing. If the component needs a tt-mlir fix, mark it as xfailed and leave a reference to the tracing issue in the test config file.

planned to add component tests along with their passing status (https://github.com/tenstorrent/tt-xla/pull/5376/) once the mlir fix is merged

PR #790, opened 2026-06-30
[diffusiongemma] Pass decoder_input_ids to skip the model's randint

kamalrajkannan78:
### Ticket

- https://github.com/tenstorrent/tt-xla/issues/5423

### Problem description

- DiffusionGemma's forward initializes the decoder canvas with [`torch.randint` when `decoder_input_ids is None`](https://github.com/huggingface/transformers/blob/e0e7504bca2bfd1b85bb0eedb148f7b250226f06/src/transformers/models/diffusion_gemma/modeling_diffusion_gemma.py#L1570-L1575). On device, `randint` lowers to `rng_bit_generator` + an unsigned `remainder` (`bits % range`), and tt-metal doesn't support `remainder` on `UINT32` → the model crashes with `Unsupported data type for remainder DataType::UINT32`.

### What's changed

- `load_inputs` now passes `decoder_input_ids` (host-side `torch.randint`, same shape/range/dtype as the model's own init), so the model skips its internal device-side `randint` and no `remainder` op is emitted. The model now compiles and runs end-to-end on 8 chips. This is just a quick workaround for the tt-metal gap (UINT32 `remainder`); removable once supported.
- With this fix and [tt-xla#5424](https://github.com/tenstorrent/tt-xla/pull/5424), models now fails at runtime with `PCC=0.96`

### Checklist
- [x] Verify the changes through local testing in WH

### Logs

- [jun30_diffgemma_before_work_around.log.zip](https://github.com/user-attachments/files/29496015/jun30_diffgemma_after_fix.log.zip)
- [jun30_diffgemma_after_work_around.log.zip](https://github.com/user-attachments/files/29516803/jun30_diffgemma_after_workaround.log.zip)

### DIFF
--- a/diffusiongemma/pytorch/loader.py
+++ b/diffusiongemma/pytorch/loader.py
@@ -11,6 +11,7 @@ MoE backbone that denoises a block of tokens instead of decoding left-to-right.
 
 from typing import Optional
 
+import torch
 from transformers import AutoProcessor
 
 from ...base import ForgeModel
@@ -106,6 +107,15 @@ class ModelLoader(ForgeModel):
         for key in list(inputs):
             value = inputs[key].repeat_interleave(batch_size, dim=0)
             inputs[key] = cast_input_to_type(value, dtype_override)
+        # Workaround: pass decoder_input_ids to skip the model's randint (its
+        # lowering hits unsupported uint32 remainder).
+        # tt-metal ticket - https://github.com/tenstorrent/tt-metal/issues/27621
+        # tt-metal pr - https://github.com/tenstorrent/tt-metal/pull/48697
+        # tt-xla tracker - https://github.com/tenstorrent/tt-xla/issues/5423
+        text_cfg = getattr(self.config, "text_config", self.config)
+        inputs["decoder_input_ids"] = torch.randint(
+            0, text_cfg.vocab_size, (batch_size, self.config.canvas_length)
+        )
         return inputs
 
     def _text_layers(self, model):


diffusiongemma/pytorch/loader.py
```diff
@@ -106,6 +107,12 @@ def load_inputs(
         for key in list(inputs):
             value = inputs[key].repeat_interleave(batch_size, dim=0)
             inputs[key] = cast_input_to_type(value, dtype_override)
+        # Workaround: pass decoder_input_ids to skip the model's randint (its
```
nvukobratTT, line 110: Why not push for an issue fix instead of doing a workaround at the input level? 

Also, is the model passing with this workaround enabled?

kamalrajkannan78: > Why not push for an issue fix instead of doing a workaround at the input level?

Agreed the proper fix belongs in tt-metal — `randint`'s lowering hits an unsupported `remainder` on UINT32 (tracked in tenstorrent/tt-metal#27621, and noted in the linked ticket). But even with a tt-metal PR, we'd still have to wait for the next uplift, and since this is a priority model I added this input-level workaround to unblock it in the meantime; it's removable once the fix lands. Since it's listed (still unchecked) in that tracker, to avoid duplicating effort I'll sync with the issue owner to confirm there's no ongoing work, and raise the PR today if it's not already being picked up.

> Also, is the model passing with this workaround enabled?

Not yet. With this workaround the model compiles and fails at runtime with a PCC=0.96, as mentioned in the description. So this clears the crash - [jun30_diffgemma_after_work_around.log.zip](https://github.com/user-attachments/files/29516803/jun30_diffgemma_after_workaround.log.zip)

nvukobratTT: Ok, instead of here, can we add MLIR workaround with link to the metal issue?

kamalrajkannan78: - Update on the proper tt-metal fix: synced with @KalaivaniMCW — @mcw-anasuya is now working on UINT32 `remainder` support (https://github.com/tenstorrent/tt-metal/issues/27621).
- This particular workaround skips the model's `randint` entirely, which can only happen at the model-input level — by the time it reaches tt-mlir, `randint` is already lowered to `rng_bit_generator` + `remainder`, so MLIR can't un-call it.
- A tt-mlir-side workaround would instead have to handle the uint32 `remainder` op itself (e.g. rewrite modulo-by-power-of-2 via shift/subtract) 
- Since Anasuya is working on proper uint32 `remainder` support in tt-metal (#27621), I'd lean toward keeping this input-level bridge (removable as soon as that lands) over a temporary MLIR rewrite. wdyt?

nvukobratTT: Ok, sounds good. Let's also track the issue link above the workaround. Approving the PR :))

nvukobratTT (approved):

PR #798, opened 2026-07-03
Add Qwen3.5 27B causal_lm loader with tensor-parallel sharding

sonalibaskaran2499:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/3508)

### Problem description
Qwen3.5 is a hybrid model (Gated DeltaNet + full attention) and needs a loader with a tensor-parallel sharding strategy. The gated-delta block's fused in_proj_qkv is sharded on the "model" axis, but the later torch.split into Q/K/V cuts that axis at boundaries that don't align with the per-device shards → Q/K/V get scrambled and PCC collapses. The conv output must be replicated before the split.

### What's changed
Adds qwen_3_5/causal_lm/pytorch/loader.py:
- Variants incl. QWEN_3_5_27B, get_mesh_config (("batch","model") mesh)
- load_shard_spec - weight TP specs (MLP, full-attn q/k/v/o, gated-delta)
- load_activation_shard_spec - declares {conv1d: None} (replicate output) so the fused-QKV split runs on correct data; applied by the tt-xla device runner

### Logs
[run_27b_full.log](https://github.com/user-attachments/files/29704108/run_27b_full.log)

### Checklist

### DIFF
--- a/qwen_3_5/causal_lm/pytorch/loader.py
+++ b/qwen_3_5/causal_lm/pytorch/loader.py
@@ -3,6 +3,11 @@
 # SPDX-License-Identifier: Apache-2.0
 """
 Qwen 3.5 model loader implementation for causal language modeling.
+
+Qwen 3.5 uses a hybrid architecture interleaving Gated DeltaNet (linear
+attention with causal conv1d + chunked delta rule) and standard full
+attention layers. Dense variants follow the layout
+(3x linear_attention + 1x full_attention) repeated.
 """
 import torch
 from transformers import AutoModelForCausalLM, AutoTokenizer, AutoConfig
@@ -21,7 +26,7 @@ from ....config import (
 
 
 class ModelVariant(StrEnum):
-    """Available Qwen 3.5 model variants for causal language modeling."""
+    """Available Qwen 3.5 dense model variants for causal language modeling."""
 
     QWEN_3_5_0_8B = "0_8B"
     QWEN_3_5_2B = "2B"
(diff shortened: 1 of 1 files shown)


qwen_3_5/causal_lm/pytorch/loader.py
```diff
@@ -169,3 +194,59 @@ def load_inputs_decode(self, dtype_override=None, batch_size=1):
...
+        The gated-delta block's fused ``in_proj_qkv`` is sharded contiguously on
+        the "model" axis; the subsequent ``torch.split`` into [Q, K, V] cuts that
+        sharded axis at points that don't align with the per-device boundaries,
+        which miscompiles under Shardy and scrambles q/k/v before the recurrence
+        (full-model PCC collapses). Replicating the conv output before the split
+        makes the split run on correct data.
+        """
+        constraints = {}
+        for layer in model.model.layers:
+            la = getattr(layer, "linear_attn", None)
```
AleksKnezevic: is there a way to not look for `linear_attn` attribute in each layer and simply select the correct ones? If so, please change, if not, that's fine too.

AleksKnezevic (approved):

qwen_3_5/causal_lm/pytorch/loader.py
sonalibaskaran2499: The loop is needed anyway. MLP weights are sharded on every layer, so the hasattr check just picks the attention weights in that same pass the layers are an interleaved hybrid so there's no pre-grouped list to select.

PR #806, opened 2026-07-09
[OmniGen] Fix transformer MLP tensor-parallel PCC drop

### LINKED ISSUE #5490: [Omnigen] Transformer Runtime Failure
`tests/torch/models/omnigen/test_transformer.py::test_transformer_sharded` fails with 
`loc("slice.1459"): error: 'sdy.collective_permute' op requires the same type for all operands and results`

meenakshiramanathan1:
### Ticket

- Fixes [#5490](https://github.com/tenstorrent/tt-xla/issues/5490)

### Problem description

Model fails with `AssertionError: Evaluation result 0 failed: PCC comparison failed. Calculated: pcc=0.009614370396655155. Required: pcc=0.99.`

### What's changed

- The strategy was designed using the sharding-model-analysis skill: the OmniGen transformer (OmniGenTransformer2DModel) architecture was mapped from the config and diffusers source, the mesh was fixed at (2,4)=8 devices, and the collective for each candidate sharding pattern was looked up in the skill's CCL cheatsheet before any code was touched.

- MLP tensor-parallel PCC drop. OmniGenFeedForward fuses gate and up into a single gate_up_proj Linear. With gate_up_proj column-parallel ("model","batch") and down_proj row-parallel ("batch","model"), the fused output dim puts the whole gate half on one device and the whole up half on another, so the subsequent up * silu(gate) needs a cross-device gather. The row-parallel down_proj (contraction dim sharded) then forces that gather to reshard back to a model-sharded layout, which the compiler lowers incorrectly and collapses PCC to ~0.1. Changed down_proj to column-parallel ("model","batch") (weight) / ("model",) (bias) so the gather resolves to a replicated intermediate — the same layout phi3 (identical fused-MLP + chunk pattern) uses successfully. Applied generically across all transformer blocks.

- Attention split unchanged. Q/K/V and gate_up_proj stay column-parallel ("model","batch") and the attention output projection stays row-parallel ("batch","model"); only down_proj moves from row- to column-parallel.

Post these changes,  pcc jumps to pcc=0.9325374453280133.

### Logs
[omnigen_transformer_before_jul14.log](https://github.com/user-attachments/files/29996544/omnigen_transformer_before_jul14.log)
[omnigen_transformer_after_jul14.log](https://github.com/user-attachments/files/29996533/omnigen_transformer_after_jul14.log)

### DIFF
--- a/omnigen/pytorch/src/model_utils.py
+++ b/omnigen/pytorch/src/model_utils.py
@@ -429,8 +429,12 @@ def shard_vae_encoder_specs(vae) -> dict:
 def shard_transformer_specs(transformer) -> dict:
     """Shard specs for OmniGenTransformer2DModel.
 
-    Column-parallel (Q, K, V, gate_up_proj): ("model", "batch")
-    Row-parallel   (attn out, down_proj):    ("batch", "model")
+    Column-parallel (Q, K, V, gate_up_proj, down_proj): ("model", "batch")
+    Row-parallel    (attn out):                         ("batch", "model")
+
+    down_proj is column-parallel (not row-parallel) on purpose: the fused
+    gate_up_proj + chunk pattern mis-lowers when down_proj shards its
+    contraction dim. See the note at the down_proj spec below.
     """
     specs = {}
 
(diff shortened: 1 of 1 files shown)


meenakshiramanathan1:
Closing since this [PR](https://github.com/tenstorrent/tt-forge-models/pull/816) will handle this issue.

PR #829, opened 2026-07-20
[diffusiongemma] Add block-diffusion text-generation pipeline

kamalrajkannan78:
### Ticket

- https://github.com/tenstorrent/tt-xla/issues/5526
- https://github.com/tenstorrent/tt-xla/issues/5538

### Problem description

- The DiffusionGemma-26B block-diffusion pipeline (the staged encoder→decoder driver) was duplicated across the demo and the PCC-gated nightly test in tt-xla — each carried its own copy of `manual_generate` + the `TTEncoder`/`TTDecoder` wrappers + the staged-residency logic - [discussion](https://github.com/tenstorrent/tt-xla/pull/5540#pullrequestreview-4733474831)

### What's changed

- Adds `diffusiongemma/pytorch/pipeline.py`, exposing `DiffusionGemmaPipeline` (`setup()` + `generate()`) plus the module-level driver (`manual_generate`), the `TTEncoder`/`TTDecoder` wrappers, and device helpers (`to_device` / `cache_to_device` / `free_tt_graphs`). The tt-xla demo imports the high-level pipeline; the nightly PCC test imports the driver + wrappers and adds only its per-step PCC staging — so neither carries its own copy.
- **Staged residency (both components on TT).** The model can't fit on device twice, so the encoder is loaded as an independent model, prefills the KV cache, is freed, then the decoder is loaded — only one component is device-resident at a time.
- **Decoder traces one graph (no per-step recompile).** In the denoising loop `self_conditioning_logits` is `None` on step 1 and a tensor after — a shape change that would retrace. The pipeline feeds a **consistent tensor every step** (zeros + `self_conditioning_mask=False` on step 1, real logits + `mask=True` after); `mask=False` exactly reproduces the `None` branch, so it stays numerically identical while the input shape never changes. Combined with `MAX_NEW_TOKENS=256` (one block → static KV cache), the decoder traces **one graph** for the whole run.
- **`TTEncoder`/`TTDecoder` wrappers** adapt the model to `torch.compile`'s tensor-I/O expectations: positional tensors in, one tensor out (`last_hidden_state` / `logits`); `TTDecoder` pins `input_ids=None` to select the decoder path.
- **Re-register `tt_moe` in `setup()`.** The caller's transformers version swap replaces `transformers` in memory, dropping `moe_backend`'s import-time `tt_moe` registration; `setup()` calls `register_tt_moe_backend()` against the live version (see tenstorrent/tt-xla#5424). The version swap itself stays on the tt-xla side (this repo can't import that test util), so the caller must install transformers ≥ 5.11 before `setup()`.

### Checklist
- [x] PCC-gated nightly test - [jul20_diff_gemma_pcc_check.log.zip](https://github.com/user-attachments/files/30209105/jul20_diff_gemma_pcc_check.log.zip) 
- [x] Demo run - [jul20_diff_gemma_demo.log.zip](https://github.com/user-attachments/files/30209091/jul20_diff_gemma_demo.log.zip)

### DIFF
--- a/diffusiongemma/pytorch/pipeline.py
+++ b/diffusiongemma/pytorch/pipeline.py
@@ -1,0 +1,442 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""DiffusionGemma 26B block-diffusion text-generation pipeline on Tenstorrent.
+
+Both the encoder (prefill) and the decoder (denoising loop) run on TT (sharded, SPMD). The
+model can't fit on device twice, so residency is STAGED: the encoder is loaded as an
+independent model, prefills the KV cache, is freed, then the decoder is loaded -- only one
+component is device-resident at a time.
+
+The host driver (sampler/stopping/cache/RNG) reuses the model's own generate() helpers, so it
+is bit-identical to generate(); only the two NN components run on TT. ``manual_generate`` and
+the ``TT*`` wrappers are module-level so a PCC test can reuse them with injected forwards.
+
+NOTE: the caller must install transformers>=5.11 (the version swap lives on the tt-xla side,
+which this repo can't import); ``setup()`` re-registers tt_moe against that swapped-in version.
+"""
+
+import copy
+import gc
+import math
+import os
+from types import SimpleNamespace
+
+import numpy as np
+import torch
+import torch_xla.core.xla_model as xm
+import torch_xla.distributed.spmd as xs
+import torch_xla.runtime as xr
+from loguru import logger
+from torch_xla.distributed.spmd import Mesh
+from tt_torch.moe_backend import TT_MOE_BACKEND_NAME, register_tt_moe_backend
+
+from .loader import ModelLoader, ModelVariant
+
+PROMPT = "Why is the sky blue?"
+MAX_NEW_TOKENS = 256  # one canvas block
+SEED = 0
+
+
+def enable_spmd():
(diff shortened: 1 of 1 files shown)


diffusiongemma/pytorch/pipeline.py
```diff
@@ -0,0 +1,437 @@
...
+        }  # decoder graph during the decode loop
+
+        def encoder_forward(**kw):
+            # Free the previous block's decoder (if any) so only one model is resident.
+            if stage["dec_tt"] is not None:
+                stage["dec_tt"] = stage["dec_model"] = None
+                free_tt_graphs()
+            enc_model = self._load_sharded(ModelVariant.ENCODER)
+            enc_tt = torch.compile(TTEncoder(enc_model), backend="tt")
+            pkv = DynamicCache()
```
sshonTT, line 368: Is this intended? Don't we need pkv from prev block? Or could we have an assertion for `max_new_canvases == 1`?

kamalrajkannan78: Good catch. It's intentional but only valid for a single canvas block — `encoder_forward` starts a fresh KV cache each block, so multi-block would drop the prior block's context (we'd need cross-block carry-over). For `MAX_NEW_TOKENS=256`, one block runs, so there's no previous block to inherit. I've added an `assert max_new_canvases == 1` in `manual_generate` to make the single-block limit explicit and fail loudly if the token budget is bumped without adding carry-over.

sshonTT (approved):

sgligorijevicTT (approved):

mrakitaTT (approved):

PR #833, opened 2026-07-22
[Qwen-Image] Add pipeline component loader + modulation sharding fix

ashokkumarkannan1:
## Ticket
- https://github.com/tenstorrent/tt-xla/issues/5753
- Root-cause context: https://github.com/tenstorrent/tt-xla/issues/5760

## Problem Description
- Qwen-Image lacked component model loaders in tt-forge-models. Deploying the pipeline component-by-component needs each segment — the Qwen2.5-VL text encoder, the QwenImage MMDiT transformer, and the VAE decoder — loadable independently with tensor wrapper interfaces and pre-configured inputs. Both the ~20B transformer and the ~7B text encoder additionally need multi-chip tensor-parallel shard specs to fit device DRAM.
- The e2e pipeline itself also belongs here, so the tt-xla demo and perf benchmark consume one implementation instead of each carrying its own copy.

### Component loaders and shard specs
- `qwen_image/pytorch` loader for `Qwen/Qwen-Image` with `ModelVariant` implementations: `TextEncoder` (Qwen2.5-VL, ~7B), `Transformer` (QwenImage MMDiT, ~20B), `Vae` (AutoencoderKLQwenImage decoder).
- **`src/model_utils.py`**: per-component loaders, tokenizer integration, tensor-in/tensor-out `nn.Module` wrappers, synthetic inputs at the native 1:1 1328×1328 resolution, and the tensor-parallel shard specs.
- **Transformer shard spec**: Megatron column→row on attention/FFN; the per-block modulation projection is sharded **contraction-parallel** (`(None, "model")`) — an all-reduce restores the full `6*dim` output before `chunk(6)`, matching the replicated case at 1/N the memory. Per-device weights drop 20.5 → 10.3 GB; a column split would straddle the `chunk(6)` boundary and is unsafe.
- **Text-encoder shard spec**: Megatron column→row on the Qwen2.5 decoder layers (q/k/v column-parallel, o row-parallel; MLP gate/up column-parallel, down row-parallel; layernorms/embeddings replicated). Drops the encoder from ~16.6 → ~4 GB/chip so it fits alongside the transformer. GQA divides cleanly at degree 4 (28 q-heads → 7, 4 kv-heads → 1).
- Inference config matches the official Qwen/Qwen-Image usage example: source example prompt + positive magic, negative prompt `" "`, 50 steps, `true_cfg_scale` 4.0, seed 42, 1:1 1328×1328.

### `pytorch/pipeline.py` — staged residency
The reusable e2e pipeline (`QwenImageConfig` / `QwenImagePipeline`, `setup()` + `generate()`, raw pixels in `[-1, 1]`, module-level `save_image()`). The diffusers `QwenImagePipeline` orchestrates; encoder and transformer are TP-sharded, the VAE is replicated.

**Every component is loaded, run and freed before the next is placed**, so peak DRAM is `max(component)` rather than their sum. Two consequences drive the rest of the design:

- **Eviction discards the compiled graph** along with the weights — the executable pins the weight buffers, so weights cannot be freed while the graph is kept. A warmup-then-time loop *across* `generate()` calls therefore measures recompilation, never warm performance.
- **Warm is measured inside one residency**: forward 1 carries the build, 2..N are cache hits, and the functional result is always forward 1's, so behaviour does not depend on `warm_iters`. The encoder gets its warm sample free from the negative-CFG prompt and the transformer from denoise steps 2..N; only the VAE, which runs once per call, needs the synthetic repeat.

Also in this PR:

- **VAE staged and freed** like the other components (reverted from keep-resident), with the singleton temporal dim sliced **in-graph** (`[:, :, 0]`). Left in, the 5D output has two singleton dims that each tile-pad to 32 — a 1024× blow-up — so at 1328×1328 bf16 a 10,581,504 B result would ask the allocator for one contiguous **10,835,460,096 B** buffer. Verified on hardware: the graph returns rank-4 `(1, 3, 1328, 1328)` and the request is **338,608,128 B**.
- **Transformer eviction fix**: `cache_context` is a *bound method* of the transformer, so it pinned the whole module through `__self__`; dropping only the compiled module left the weights resident and eviction was silently a no-op. Both references are now nulled, and eviction is verified by `weakref` rather than assumed.
- **Per-stage compile-time and graph counters**, so cost is attributed from counters instead of assumed. A warm forward must add zero graphs.
- **`TEXT_ENCODER_CLS` / `DENOISER_CLS` / `VAE_CLS` seams**, so a test can subclass the wrappers to add PCC checks without duplicating the staging logic.

## Validation Results
4-chip Blackhole (`TT_VISIBLE_DEVICES=0,1,2,3`), via the tt-xla tests in #5752.

| Run | Result |
|---|---|
| `test_qwen_image_pipeline` (50-step nightly, PCC-gated) | ✅ passed — 3372s (55m16s) |
| `test_qwen_image` (benchmark) | ✅ passed — 2076s (33m27s) |
| demo `examples/pytorch/qwen_image.py` | ✅ 1328×1328 image — 2096s (33m50s) |

Component PCC (TT vs CPU, bf16), all gated at ≥ 0.99:

| Component | PCC |
|---|---|
| Text encoder | 1.000000 |
| Transformer (4-chip TP-sharded) | 0.996094 |
| VAE decoder | 1.000000 |

Cold vs warm, measured inside each residency:

| Component | Cold (carries the build) | Warm |
|---|---|---|
| `text_encoder` | 55.150s | 33.222s |
| `transformer_step` | 389.563s | **24.663s** |
| `vae` | 60.701s | 23.632s |

Warm reproduced across **three runs and two kernel-cache states** — `transformer_step` 24.663 / 24.656 / 24.702s, a 0.03% spread — confirming warm is residency-local and cache-independent. The falsifiability check holds every run: `warm_steps: compile 0.000s, 0 graph(s)`.

Compile is not the bottleneck: 11.753s / 4 graphs (encoder) and 46.499s / 3 graphs (transformer+VAE), i.e. **3.1% of wall**.

## Known limitations (follow-up, not addressed here)
- **A second `generate()` call in one process OOMs.** Eviction is verified host-side by `weakref`, which cannot see device buffers; ~21.7 GiB remains allocated across calls and the second VAE decode fails by 18.8 MiB on contiguity. Nothing in CI takes this path — the benchmark, demo and PCC test each make exactly one call — but any looping use is affected. DiffusionGemma's `free_tt_graphs()` (tenstorrent/tt-forge-models#891) resets dynamo and nulls the torch_xla `GraphInputMatcher` tensors, which is the likely fix.
- Relatedly, the staged design recompiles every component per call, so a looping process hits `torch._dynamo` `recompile_limit` (default 8) and silently falls back to eager.

## Checklist
- [x] New/existing tests provide coverage (validated through the Qwen-Image tests in #5752)
- [x] Demo and benchmark consume one shared pipeline implementation
- [x] Eviction verified on hardware, not assumed
- [x] Warm numbers falsifiable via graph counters

## Logs
- [qwen_benchmark.log](https://github.com/user-attachments/files/30304588/qwen_benchmark.log)
- [qwen_demo.log](https://github.com/user-attachments/files/30304589/qwen_demo.log)
- [qwen_nightly_pipeline.log](https://github.com/user-attachments/files/30304590/qwen_nightly_pipeline.log)


kamalrajkannan78 (approved):

kamalrajkannan78:
Let's wait for these PRs to land to avoid structural rewrites

[[Hunyuan_image_2_1] Add HunyuanImage 2.1 text-to-image pipeline](https://github.com/tenstorrent/tt-forge-models/pull/827)
[[diffusiongemma] Add block-diffusion text-generation pipeline](https://github.com/tenstorrent/tt-forge-models/pull/829)
[[HunyuanImage-2.1] Add PCC gated nightly test, benchmark, and demo](https://github.com/tenstorrent/tt-xla/pull/5697)
[[diffusiongemma-26B-A4B-it] Add text-to-text e2e PCC test + demo](https://github.com/tenstorrent/tt-xla/pull/5540)

kamalrajkannan78 (approved):

kmabeeTT (approved):

nvukobratTT (approved):

PR #848, opened 2026-07-30
[Hunyuan_image_2_1] Run text encoders on TT

kamalrajkannan78:
### Ticket

- https://github.com/tenstorrent/tt-xla/issues/5847

### Problem description

- HunyuanImage 2.1 only ran the transformer on TT; Qwen and ByT5 text encoders stayed on CPU.

### What's changed

- Moved Qwen (TP-sharded) and ByT5 (replicated) text encoders onto TT in `HunyuanImage21Pipeline`; scheduler and VAE stay on CPU in fp32.
- Switched the three TT components to bf16 (`TT_DTYPE`). In fp32 the encoders and the transformer together exceed a chip's DRAM, so they could not be co-resident; in bf16 the three fit with room to spare. The host-side trajectory (latents, scheduler, guidance) stays fp32 and casts at the device boundary.
- Sharded Qwen even though it is only 8B: without a shard spec, SPMD replicates it onto every chip, which crowds out the transformer.
- Loaded, compiled and uploaded each TT component once in `setup()` and left it resident, so `generate()` is pure compute and repeat calls reuse both the weights and the compiled graphs.
- Compiled `module.forward` rather than the module, so each attribute stays an `nn.Module` and callers can wrap the forward (the nightly PCC check does).
- Added `QwenPromptEmbedsWrapper` so Qwen's TT forward emits only prompt embeds, not all hidden states.

### Checklist

- [x] Demo - [demo.zip](https://github.com/user-attachments/files/30632583/demo.zip)
- [x] Benchmark - [benchmark.zip](https://github.com/user-attachments/files/30632586/benchmark.zip)
- [x] PCC-gated Nightly test - [pcc_check.zip](https://github.com/user-attachments/files/30632585/pcc.zip)

### Generated Image

Prompt :

```
"A cute, cartoon-style anthropomorphic penguin plush toy with fluffy fur, "
"standing in a painting studio, wearing a red knitted scarf and a red beret "
"with the word 'Tencent' on it, holding a paintbrush with a focused "
"expression as it paints an oil painting of the Mona Lisa, rendered in a "
"photorealistic photographic style."
```
hunyuan_image_2_1_output

### DIFF
--- a/hunyuan_image_2_1/pytorch/pipeline.py
+++ b/hunyuan_image_2_1/pytorch/pipeline.py
@@ -3,20 +3,22 @@
 # SPDX-License-Identifier: Apache-2.0
 """HunyuanImage 2.1 (Distilled) text-to-image pipeline running on Tenstorrent.
 
-Only the MMDiT transformer (the heavy net, ~17.45B) runs on the TT backend —
-tensor-parallel sharded across the device mesh and compiled with
-``torch.compile(backend="tt")``. The Qwen2.5-VL and ByT5 text encoders, the
-scheduler, and the VAE run on CPU in fp32. This mirrors the SD3 / Flux pipelines:
-precision-sensitive encoding + the VAE stay on CPU while the dominant compute is
-offloaded to TT.
+The Qwen2.5-VL encoder, the ByT5 glyph encoder and the MMDiT transformer run on
+TT in bf16; the scheduler and the VAE run on CPU in fp32. Qwen and the
+transformer are tensor-parallel sharded; ByT5 (0.22B) carries no shard spec, so
+SPMD replicates it across the mesh.
+
+Every TT model is loaded, compiled and uploaded once in ``setup()`` and stays
+resident: in bf16 the three together are ~13 GB of a chip's 32 GB. Repeat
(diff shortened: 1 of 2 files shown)


nvukobratTT (commented):
Do we know from where we're getting this noise?

kamalrajkannan78:
> Do we know from where we're getting this noise?

In this model the text encoders and the transformer run on TT. Their PCC values are:

| Component | Fwd | PCC |
|---|---|---|
| text_encoder (Qwen, sharded, bf16) | 1 | 0.949 |
| text_encoder_2 (ByT5, bf16) | 1 | 1.000 |
| transformer (sharded, bf16) | 1–8 | 0.992, 0.934, 0.906, 0.918, 0.930, 0.949, 0.945, 0.910 |

Since PCC is above 0.90 for every component's forward runs, I went ahead with the demo. The PCC drop is most likely the reason for the noise in the final image — I'll investigate this further and address it as follow-up work. 

Log - [pcc_check.zip](https://github.com/user-attachments/files/30632585/pcc.zip)

nvukobratTT (approved):

PR #849, opened 2026-07-31
[gemma4] Strip mask-gather assertions and pin transformers for multimodal paths

saiarthiraguram:
### Problem description
#745 added the `12B-image` / `12B-audio` / `12B-video` variants to the gemma4
loader, but those paths do not actually run on device.

`Gemma4UnifiedModel.forward` guards each modality merge with a shape assertion:

```python
torch_compilable_check(
    inputs_embeds[image_mask].numel() == image_features.numel(), ...)
```

`image_mask` has already been `expand_as(inputs_embeds)`-ed, so
`inputs_embeds[image_mask]` is a boolean-mask advanced index over the whole
`(batch, seq, hidden)` embedding. Its result shape is data dependent, so
TorchDynamo graph-breaks, lowers the index to a **per-scalar gather**, and ships
the entire result back to host just to read `.numel()`. On an n300-llmbox that
gather is the whole failure:

| modality | gather | failure |
|---|---|---|
| image | 266×3840 = 1,021,440 scalars | bf16→f32 typecast in the index chain lands width-sharded in L1 at 499 tiles/core — 2,043,904 B/bank against a 1,328,064 B bank → `TT_FATAL: Out of Memory`, 130,809,856 B L1 buffer |
| audio | same shape of failure | 72,351,744 B L1 buffer |
| video | 2016×3840 = 7,741,440 scalars | `EmbeddingOp`'s `TilizeWithValPadding` asks for a 15,854,469,120 B DRAM buffer |

Two smaller issues in the same paths: the video clip was built by replicating one
still frame, and the processor was given no frame metadata.

### What's changed

**`_strip_mask_gather_checks()`** — AST-rewrites `Gemma4UnifiedModel.forward` to
remove the three merge assertions before anything traces it, then recompiles the
function and registers the rewritten source with `linecache` so
`inspect.getsource` keeps working (both transformers and dynamo introspect it).
The assertions are pure shape checks that never touch the tensors `forward`
returns, so removing them is numerically a no-op. It also drops three graph
breaks: the failure message is an f-string interpolating `mask.sum()`, which
forces a host sync of its own.

`TRANSFORMERS_DISABLE_TORCH_CHECK=1` is **not** sufficient — it only makes
`torch_compilable_check` return early, so Python still evaluates the
`inputs_embeds[mask].numel()` argument and the gather (and the OOM) remain. The
check has to disappear from the traced source, not just be disabled.

The strip is idempotent, and is called from `load_model()` so it covers every
variant (harmless on the text path, where the checks live inside the
per-modality branches).

**Video input fidelity** — the clip is now synthesized by panning the sample
image via `np.roll` rather than stacking one frame `T` times. Stacking is
degenerate for this model: Gemma4 attends bidirectionally across the whole video
block, so identical frames give near-tied attention logits that bf16 and fp32
resolve differently. On CPU that alone drops a bf16-vs-fp32 forward to PCC
**0.977**, against **0.991** for a panning clip — noise the device path then
inherits. A pan is also closer to real video than a freeze-frame. Frame count,
patch count and video token count are unchanged.

An explicit `VideoMetadata` at `VIDEO_FPS = 1.0` is now passed as well. Gemma4
interpolates a per-frame `mm:ss` timestamp into the prompt; with pre-sampled
frames and no metadata the processor cannot infer the frame rate and silently
falls back to `fps=24`, making the prompt depend on a library default. Only the
timestamp text depends on this — the video token count does not.

**`gemma4/pytorch/requirements.txt`** (new) — pins `transformers==5.10.4`, the
version the multimodal paths were brought up and PCC-verified against, and the
one whose assertion shape the strip is written to match. gemma4 previously had
no per-loader requirements and ran on whatever transformers happened to be
installed; the pin also protects it from sibling loaders that move the
environment mid-session (e.g. `diffusiongemma` pins `transformers==5.12.0`).

**Warn instead of silently no-oping** — if the AST match finds a number of
assertions other than the expected 3 (one per modality), the loader now raises a
`RuntimeWarning` naming the running transformers version and pointing at the
pin. Previously a restructure upstream would make the strip a quiet no-op and
the multimodal paths would resurface as unexplained device OOMs; a partial match
is warned about too, since that means one modality silently lost its workaround.

### Checklist
- [x] Strip verified against the pinned transformers 5.10.4: applies with zero
      warnings, `inspect.getsource` still works on the patched forward, 0
      remaining `inputs_embeds[...]` checks.
- [x] Warning path verified by simulating an upstream restructure (guard
      renamed so nothing matches): fires `RuntimeWarning`, no crash, original
      forward left intact.
- [x] Exercised by tt-xla `tests/benchmark/test_multimodal.py`
      (`test_gemma4_12b_{text,image,audio,video}`, companion PR) on
      n300-llmbox.
      
 ## logs
 
[gemma4_12b_audio_fix1.log](https://github.com/user-attachments/files/30584703/gemma4_12b_audio_fix1.log)
[gemma4_12b_image_fix1.log](https://github.com/user-attachments/files/30584704/gemma4_12b_image_fix1.log)
[gemma4_12b_text.log](https://github.com/user-attachments/files/30584705/gemma4_12b_text.log)
[gemma4_12b_video_pan.log](https://github.com/user-attachments/files/30584706/gemma4_12b_video_pan.log)


gemma4/pytorch/requirements.txt
```diff
@@ -0,0 +1,8 @@
+# google/gemma-4-12B is a Gemma4UnifiedForConditionalGeneration, which lives in
+# transformers.models.gemma4_unified. The multimodal (image / audio / video)
+# paths were brought up and PCC-verified against 5.10.4, and the loader's
+# _strip_mask_gather_checks() AST-matches the three torch_compilable_check
+# merge assertions as they are written in that release. Pin it: on a version
+# where those checks are restructured the strip degrades to a no-op and the
+# multimodal paths OOM (130 MB L1 image / 72 MB L1 audio / 15.8 GB DRAM video).
+transformers==5.10.4
```
nvukobratTT, line 8: Is this required? Can we solve the problem in the current version of the transformers lib? :))

saiarthiraguram: <head></head><h3 dir="auto" data-pm-slice="0 0 []" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code dir="ltr">5.10.1</code><span class="Apple-converted-space"> </span>Is a Required Minimum Version as model does not exist below it</h3><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">The minimum version is required because the model is not available in earlier Transformers releases.</p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"></p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code dir="ltr">google/gemma-4-12B</code><span class="Apple-converted-space"> </span>is a<span class="Apple-converted-space"> </span><code dir="ltr">Gemma4UnifiedForConditionalGeneration</code>, implemented under:</p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"></p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code dir="ltr">transformers.models.gemma4_unified</code></p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"></p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">This module was first introduced in<span class="Apple-converted-space"> </span><strong>Transformers 5.10.1</strong>.</p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"></p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">The shared test environment currently uses<span class="Apple-converted-space"> </span><code dir="ltr">transformers==5.5.1</code><span class="Apple-converted-space"> </span>through<span class="Apple-converted-space"> </span><code dir="ltr">tt-xla/venv/requirements-dev.txt</code>. While that version contains<span class="Apple-converted-space"> </span><code dir="ltr">transformers.models.gemma4</code>, it does not contain<span class="Apple-converted-space"> </span><code dir="ltr">gemma4_unified</code>, so<span class="Apple-converted-space"> </span><code dir="ltr">AutoConfig</code><span class="Apple-converted-space"> </span>cannot resolve the checkpoint's<span class="Apple-converted-space"> </span><code dir="ltr">model_type</code>.</p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"></p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">In other words, this cannot be addressed at the loader level because the required model implementation simply does not exist in the installed release.</p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"></p>

<p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Therefore, the minimum version should remain at<span class="Apple-converted-space"> </span><strong>5.10.1</strong>.</p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"></p><p dir="auto" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Once the repo-wide Transformers dependency is updated to<span class="Apple-converted-space"> </span><code dir="ltr">&gt;=5.10.1</code>, this compatibility workaround can be removed entirely. I have also documented this directly in the file so that it is clear why the dependency floor exists.</p>

PR #855, opened 2026-08-10
Bringup Molmo Model

devisettymahidhar608:
### Provide context for the problem

Adds bringup for **AllenAI Molmo-72B-0924**  as a
tensor-parallel inference model.

Molmo's decoder MLP uses a **fused SwiGLU projection**: a single `ff_proj`
Linear (`8192 -> 59136`) whose output is logically two stacked halves
`[ up(0:29568) | gate(29568:59136) ]`, later split by `SwiGLU.chunk(2, dim=-1)`
into `silu(gate) * up`.

Column-sharding that fused weight across the mesh places the `up` half on one
set of devices and the `gate` half on another. Because this flow lowers a
`chunk`/`split` on a sharded axis **locally per shard** (no correctness-preserving
re-gather), each device's `chunk` mis-pairs `up`/`gate`. Repeated across all 80
layers this destroys the output — tensor-parallel inference PCC collapsed to
**~0.028** (required 0.99).

### Describe the approach used to solve the problem

Unfuse `ff_proj` into two independent `up` / `gate` Linears at load time
(`_split_molmo_ff_proj` / `_SplitSwiGLUFeedForwardProj`), matching how every
other dense SwiGLU model in the repo is already structured, and the repo's
existing "unfuse fused projections" pattern (`infinity`, `gpt_oss`). Each
projection is then column-sharded independently on the **same** output
dimension, so every device holds matching `up[i]` and `gate[i]` and the SwiGLU
multiply is a correct local element-wise op; the row-parallel `ff_out` consumes
the matching shard.

Key properties:
- **Numerically identical** to the fused path (`silu(gate(x)) * up(x)` ==
  `SwiGLU(ff_proj(x))`; verified max abs diff `~5e-7`), so single-device runs
  are unchanged.
- **Shard-count-agnostic** — never relies on split boundaries aligning with
  shard boundaries.
- **No memory regression** — the largest weight in the model stays sharded.

The fused GQA `att_proj` is left replicated (same fused-split hazard, unequal
Q/K/V widths), consistent with the existing loader comment.

**Result:** `molmo/pytorch-Molmo-72B-0924-tensor_parallel-inference` PCC
0.028 -> >= 0.99 (passing).

logs:
[molmo.log](https://github.com/user-attachments/files/30892054/molmo.log)


molmo/causal_lm/pytorch/loader.py
```diff
@@ -0,0 +1,415 @@
...
+        """Load tokenizer for the current variant.
+        Args:
+            dtype_override: Optional torch.dtype to override the tokenizer's default dtype.
+
+        Returns:
+            The loaded tokenizer instance
+        """
+        tokenizer_kwargs = {}
+        if dtype_override is not None:
+            tokenizer_kwargs["torch_dtype"] = dtype_override
```
kmabeeTT: I think this could be removed, was removedin many other models in the past, I flagged this on some of @sonalibaskaran2499 PR's recently

kmabeeTT (approved):

molmo/causal_lm/pytorch/loader.py
devisettymahidhar608: thanks kyle, i will remove.

PR #884, opened 2026-08-19
[Ideogram 4] Run the Qwen3-VL text encoder on TT

### LINKED ISSUE #5981: [Ideogram 4] Run the Qwen3-VL text encoder on TT
Ideogram 4 conditions its DiT on `llm_features` — hidden states tapped from 13 of the Qwen3-VL language tower's 36 decoder layers, interleaved along the feature axis into `(B, L, 4096*13)`. The `ideogram_4/pytorch` loader exposes the DiT (`Transformer_FP8_512`) and the `Vae` decoder, but not that tower, so it has no test id and nothing can compile or measure it on device. It is the model's last compute component still on host.

Sibling work on this loader: the DiT tensor-parallel shard specs (tenstorrent/tt-forge-models#859), the VAE decoder component (tenstorrent/tt-forge-models#860), and the e2e pipeline (tenstorrent/tt-forge-models#872). Mirrors the component split done for FIBO's SmolLM3 text encoder (tenstorrent/tt-forge-models#874). Part of #5171, which stays open for the e2e pipeline.

**Scope**

- A `TextEncoder` variant on the `ideogram_4` loader (7.568B params, 14.1 GiB bf16) whose wrapper reproduces `Ideogram4Pipeline._encode_text` as a drop-in: same three input tensors, one returned fp32 tensor of tap-interleaved hidden states masked to the LLM indicator positions — what the pipeline hands the DiT as `llm_features`, not the tower's `last_hidden_state`.
- Construct only the language tower. `Qwen3VLModel` would also build the vision tower, which text-only prompts never reach, so it would materialize and place ~0.5B unused parameters on device.
- Materialize the published weight-only FP8 modules to plain bf16 `nn.Linear` via the existing `materialize_fp8_state_dict_to_bf16`, as `load_conditional_transformer` already does for the DiT.
- Single-device and unsharded: `get_mesh_config` returns `(1, 1)` and `load_shard_spec` returns `None`. Qwen3-VL is cleanly shardable (32 q / 8 kv heads and a 12288 MLP all divide 2, 4 and 8), unlike this DiT's fused 18-head attention, but that cannot be validated on the single chip used for bringup.
- CPU-vs-TT validation on a Wormhole n150 with real weights, plus before/after generated images with the encoder as the only variable.
- A tt-xla runner entry `ideogram_4/pytorch-TextEncoder-single_device-inference` (`EXPECTED_PASSING`, `enable_weight_bfp8_conversion: true`), following the submodule uplift alongside the still-pending `Vae` entry.

**PRs**

- tenstorrent/tt-forge-models#884

ctr-lelanchelianTT:
### Ticket
Closes tenstorrent/tt-xla#5981.

Part of https://github.com/tenstorrent/tt-xla/issues/5171 — does not close it; the
e2e pipeline (#872) is still draft. Sibling work on this loader: #859 (DiT
tensor-parallel shard specs), #860 (VAE decoder component), #872 (e2e pipeline).
Mirrors the component split in #874 (FIBO's SmolLM3 text encoder).

### Problem description
Ideogram 4 conditions its DiT on `llm_features` — hidden states tapped from 13 of
the Qwen3-VL language tower's 36 decoder layers, interleaved along the feature axis
into `(B, L, 4096*13)`. The loader exposed the DiT (`Transformer_FP8_512`) and the
`Vae` decoder, but not that tower, so it had no test id and nothing could compile or
measure it on device. It was the model's last compute component still on host.

### What's changed

**`TextEncoder` variant** on the `ideogram_4` loader (7.568B params, 14.1 GiB bf16).

`Ideogram4TextEncoderWrapper.forward(token_ids, text_position_ids, indicator)`
reproduces `Ideogram4Pipeline._encode_text` — and the `_get_qwen3_vl_embeddings` it
calls — so the component is a drop-in for that method: same three input tensors,
same single returned tensor. It returns **one** fp32 tensor `(1, 1280, 53248)`, the
tap-interleaved hidden states masked to the `LLM_TOKEN_INDICATOR` positions, which
is exactly what the pipeline hands the DiT as `llm_features` (not a tuple, and not
the tower's `last_hidden_state`).

`load_inputs` returns three int64 tensors: `token_ids (1, 1280)`,
`text_position_ids (1, 1280, 3)`, `indicator (1, 1280)`. They are built from the
real tokenizer over `DEFAULT_JSON_CAPTION`, left-padded to `MAX_TEXT_TOKENS=256` and
followed by `NUM_IMAGE_TOKENS=1024` image slots, giving the same `TOTAL_SEQ_LEN=1280`
the DiT component uses. That padded layout is what `_build_inputs` itself produces
whenever the longest prompt in the batch is 256 tokens, so nothing is synthetic
beyond the choice of prompt.

**Only the language tower is constructed.** `Qwen3VLModel` would also build the
vision tower; text-only prompts never reach it, so that would materialize and place
~0.5B unused parameters on device. Its tensors still arrive in the same safetensors
shards and are dropped before load.

**Weights are materialized to plain bf16 `nn.Linear`** via the existing
`materialize_fp8_state_dict_to_bf16`, the same way `load_conditional_transformer`
handles the DiT, rather than kept as the published weight-only FP8 modules that
dequantize inside every matmul. Loaded `strict=True`.

**No shard spec.** `get_mesh_config` returns `(1, 1)` for any device count and
`load_shard_spec` returns `None` — single-device, unsharded. Qwen3-VL is in fact
cleanly shardable (32 q / 8 kv heads and a 12288 MLP all divide 2, 4 and 8), unlike
this DiT's fused 18-head attention, but that could not be validated on the single
chip this component was brought up on.

`DEFAULT_VARIANT` stays `Transformer_FP8_512` and both existing variants are
untouched, so the existing test ids keep resolving.

### Verification
Single Wormhole n150, real weights, full 36-layer tower, all 13 taps.

| Check | Result |
|---|---|
| Component vs upstream `_encode_text`, CPU | **bit-exact** (`torch.equal`) |
| bf16-materialized vs published FP8 weights, CPU | PCC **0.999958** |
| TT vs CPU golden, `experimental_weight_dtype=bfp_bf8` | PCC **0.999603** |
| Compile + first forward | 95–104 s |
| Runner `ideogram_4/pytorch-TextEncoder-single_device-inference` | **PASSED** in 442 s |

bfp8 weight conversion is required for the same reason as the DiT's runner entry:
14.1 GiB of bf16 weights does not fit a single chip's DRAM.

Per-tap PCC degrades monotonically with depth — L0 0.99990, L3 0.99981, L6–L21
0.99995–0.99998, L24 0.99989, L27 0.99979, L30 0.99960, L33 0.99940, L35 0.99797 —
consistent with bfp8 error accumulating over 36 layers rather than any single bad
op. The large max absolute diff (832) is Qwen's massive-activation outliers, not a
correctness problem: golden max abs is 1.27e4 against a mean of 2.14, so it is
1.68% mean relative error across the 1699 elements above 100, and PCC 0.998613 on
everything at or below 100.

**e2e image, encoder as the only variable.** The same DiT and VAE run on CPU in both
images, fed `llm_features` computed either on host or on TT — 512x512, 12 steps,
guidance 4.0, seed 1234, JSON caption. Both are coherent and prompt-faithful; **image
RGB PCC 0.946395, MAE 12.12/255**. The residual is sampler-trajectory divergence rather
than encoder disagreement: the encoder feeds the denoising loop, so a 4e-4 feature
difference nudges the trajectory over 12 steps — the same effect seen on FIBO's text
encoder in #874. A fully-on-device e2e is not possible on the chip available, since the
9.3B DiT and the 14.1 GiB encoder cannot co-reside.

**MRoPE, measured rather than assumed.** transformers'
`Qwen3VLTextRotaryEmbedding.apply_interleaved_mrope` writes the H and W axes into a
view of `freqs` with an in-place strided scatter — the same construct that had to be
rewritten in this DiT's own MRoPE, where it mislowered into a position-correlated
cos/sin error. **It does not need rewriting here.** Ablated on device: upstream
scatter PCC 0.999601 vs an equivalent masked sum 0.999603 (bulk 0.998571 vs
0.998613). The DiT is sensitive because rope is recomputed from the same
`position_ids` at every denoising step, so a small position-correlated error
resonates; the text encoder runs once per generation, so it cannot. No patch is
shipped — a global monkey-patch of a transformers class would leak into any other
Qwen3-VL loader in the same process for 2e-6 of PCC. The finding and both numbers
are recorded as a comment in `load_text_encoder`.

### Land order
No submodule bump in this PR. The tt-xla runner entry
(`ideogram_4/pytorch-TextEncoder-single_device-inference`, `EXPECTED_PASSING` with
`enable_weight_bfp8_conversion: true`) follows the uplift, alongside the still-pending
`Vae` entry from #860.

### Checklist


nvukobratTT (approved):
Let's just to remove the image artefacts.

PR #900, opened 2026-08-25
[pyramid_flow] Run the CausalVideoVAE decoder on TT

### LINKED ISSUE #6006: [pyramid_flow] Run the CausalVideoVAE decoder on TT
The `pyramid_flow/pytorch` loader exposes the miniFLUX DiT and the CLIP text encoder, but its own docstring records the rest of the state plainly: *"the CausalVideoVAE and flow-matching scheduler have not been brought up. Those remain CPU-side."* The VAE is therefore Pyramid Flow's last component with no TT path at all.

Two consequences. Every latent the model generates has to be decoded to pixels on the host, so the decode is charged to CPU in every end-to-end run. More limiting for bringup work, the VAE has no runner test id, so nothing can compile it, measure it, or catch a regression in it on device.

Sibling work on this loader: the CLIP text encoder plus the SPMD text-ids lowering fix (tenstorrent/tt-forge-models#841). Mirrors the video-VAE component split done for FIBO's Wan 2.2 decoder (tenstorrent/tt-forge-models#875). Part of #4447, which stays open for the rest of the model.

**Scope**

- A `Vae` variant on the `pyramid_flow` loader: `CausalVideoVAE` decode, latents to pixels, 225.77M (decoder + `post_quant_conv`).

  | Item | Detail |
  |---|---|
  | Source | `rain1011/pyramid-flow-miniflux`, subfolder `causal_video_vae`, snapshot `c0ea661a` |
  | Architecture | Causal 3D conv autoencoder; 4 up blocks, channels 128/256/512/512, 3 layers per block, 32 norm groups, SiLU, spatial+temporal upsample `(True, True, True, False)` |
  | Latent -> output | `(1, 16, 1, 32, 32)` -> `(1, 3, 1, 256, 256)`; 16 latent channels, 8x spatial |

- `CausalVideoVAE` has no diffusers integration, so the decode path is vendored under `src/video_vae/` on the same terms as the DiT in `src/flux_modules/`: upstream code with the context-parallel `utils` imports replaced by a local stub, SPDX headers, and the repo's black formatting. Only what `decode` needs is vendored; upstream's training-only loss, LPIPS and discriminator modules are left out.
- The wrapper compiles `vae.decode(...)`, not the bare `decoder` submodule, because `PyramidDiTForVideoGeneration.decode_latent` calls `decode`. The encoder tower and `quant_conv` are dropped after load, since decode never touches them.
- The single-shot `temporal_chunk=False` path is used. Upstream's chunked decode walks temporal windows in a Python loop carrying causal-conv state between iterations and does not trace; at one latent frame that loop degenerates to a single window, and the two paths must be shown **bit-exact** rather than assumed equivalent.
- `load_inputs` draws the latent in fp32 and casts. `torch.randn` consumes its generator differently per dtype, so a seeded bf16 draw is unrelated to the fp32 draw from the same seed — drawing at the requested dtype makes a correct decode measure a meaningless PCC against an fp32 reference.
- `DEFAULT_VARIANT` and the two existing variants stay unchanged, so existing test ids keep resolving.
- CPU-vs-TT validation on a Wormhole n150, and one tt-xla runner entry, `pyramid_flow/pytorch-Vae-single_device-inference` (`EXPECTED_PASSING`), following the submodule uplift.

**Out of scope**

- **One latent frame only.** The multi-frame chunked decode still does not trace, so this covers `T=1` — the shape the image branch decodes and the shape a video decode's first window takes, but not the whole video path.
- No end-to-end video: the existing CPU-vs-TT video comparison is separate evidence and unchanged by this work.
- Single-device only; this loader carries no shard specs for any variant.
- The other two gaps stay open and stay recorded in the loader docstring: the T5-XXL encoder (runs on device, PCC 0.86) and the flow-matching scheduler.

**PRs**

- tenstorrent/tt-forge-models#900

ctr-lelanchelianTT:
### Ticket
closes https://github.com/tenstorrent/tt-xla/issues/6006

TEN-1066 — https://multicoreware.atlassian.net/browse/TEN-1066 Sibling work:
#841 (the CLIP text encoder + the SPMD text-ids lowering fix). The loader docstring
recorded this component as not brought up; this PR removes that line.

### Problem description
The `pyramid_flow` loader exposed the miniFLUX DiT and the CLIP text encoder, and
its docstring stated plainly that "the CausalVideoVAE and flow-matching scheduler
have not been brought up." So every latent Pyramid Flow generated had to be decoded
to pixels on the host, and the VAE had no test id — nothing could compile or measure
it on device. It was the model's last component with no TT path at all.

### What's changed
A `Vae` variant on the `pyramid_flow` loader (225.77M) that decodes latents to
pixels on device.

`CausalVideoVAE` has no diffusers integration, so its decode path is vendored in
`src/video_vae/` on exactly the terms the DiT already uses in `src/flux_modules/`:
upstream code, with the context-parallel `utils` imports replaced by a local
`_cp_stub.py` (mirroring the existing `_sp_stub.py`), an SPDX header, and the
reformatting the repo's black hook applies. Only what `decode` needs is vendored —
upstream's `__init__` also exports `LPIPSWithDiscriminator` and
`CausalVideoVAELossWrapper`, which pull in the training-only loss, LPIPS and
discriminator modules; those are left out.

`CausalVaeDecoderWrapper.forward` returns
`vae.decode(latents, is_init_image=True, temporal_chunk=False, return_dict=False)[0]`
— a plain pixel tensor. Two choices there are worth stating:

- **`decode`, not the bare `decoder` submodule**, because
  `PyramidDiTForVideoGeneration.decode_latent` calls `self.vae.decode(...)`, so that
  is the unit the pipeline actually runs.
- **The single-shot path, not upstream's chunked one.** `decode_latent` passes
  `temporal_chunk=True`, which slices the latent into temporal windows and walks
  them in a Python loop, decoding each with `is_init_image=False` so the causal
  convs carry state between iterations — that loop does not trace. At one latent
  frame it degenerates: `chunk_decode` builds a single-element `frame_list`
  (`full_chunk_size` goes negative, so the loop body never runs) and decodes it with
  `is_init_image=True`, which is what the wrapper does directly. **The two are
  bit-exact at `T=1`** (`torch.equal` true, max abs diff 0.0), so the traceable path
  is the same math the pipeline runs for a single frame, not an approximation of it.
  This is the same T=1 argument as FIBO's `WanVaeDecoderWrapper` in #875, with the
  equivalence measured rather than reasoned.

The encoder tower and `quant_conv` are dropped after load: `decode` never touches
them, and keeping them would roughly double the resident parameters of a
decode-only component.

`DEFAULT_VARIANT` is unchanged (`miniFLUX_768p`), the existing two variants are
untouched, and `_get_model_info`'s task ternary already routes everything but
`ClipTextEncoder` to `MM_VIDEO_TTT`, so it needed no change. No shard specs: this
loader has none for any variant, and one chip is where this was validated — a mesh
spec that cannot be exercised is not worth asserting.

`load_inputs` draws the latent **in fp32 and casts**, rather than drawing at the
requested dtype. `torch.randn` consumes the generator differently per dtype, so a
seeded bf16 draw is unrelated to the fp32 draw from the same seed — measured PCC
**0.0098** between them. Drawing directly at bf16 made a correct decode measure PCC
0.034 against an fp32 reference: two different latents, not a lowering bug. Casting
keeps the latent dtype-independent so a CPU fp32 reference and a bf16 device run
decode the same values.

### Verification
Single Wormhole **n150**, `ARCH_NAME=wormhole_b0`. Latent `(1, 16, 1, 32, 32)` ->
frame `(1, 3, 1, 256, 256)`.

**CPU-vs-TT accuracy**, CPU fp32 golden vs TT bf16, both driven through the loader's
own `load_model` / `load_inputs`:

| Tensor | PCC | MAE | max abs diff | golden max abs |
|---|---:|---:|---:|---:|
| decoded frame | **0.999956** | 0.005481 | 0.059273 | 2.093321 |

Clears the default 0.99 gate. CPU decode 3.3 s; TT compile + first forward 25-28 s
across runs.

**Upstream equivalence.** `decode(temporal_chunk=False)` versus the exact call
`decode_latent` makes (`temporal_chunk=True, window_size=1,
tile_sample_min_size=256`) on the same latent: **bit-exact**, max abs diff 0.0. This
checks the wrapper contract, not just the numerics.

**Runner, before -> after.** Before, the model exposes 12 `pyramid_flow` ids (two
variants x six configs) and no VAE id — asking for one gives `no tests collected`.
After:

```
tests/runner/test_models.py::test_all_models_torch[pyramid_flow/pytorch-Vae-single_device-inference] PASSED
================== 1 passed, 25 warnings in 106.73s (0:01:46) ==================
```

Conv3d lowers on this path — `prepare_conv3d_weight` and the padded 5-D convolutions
are visible in the TTIR/TTNN dumps.

### Scope limits, stated plainly
- **One latent frame.** The multi-frame chunked decode still does not trace (the
  stateful Python loop above), so this component covers `T=1`. That is the shape the
  384p/768p image branch decodes, and the shape a video decode's first window takes,
  but it is not the whole video path.
- **No end-to-end video in this PR.** The e2e CPU-vs-TT video comparison is separate
  evidence and unchanged by this PR; here the VAE is validated standalone and against
  the upstream reassembly.
- **Single-device only**, per the shard-spec note above.
- The other two gaps in the loader docstring are untouched and still recorded there:
  the T5-XXL encoder (runs on device, PCC 0.86) and the flow-matching scheduler.

### Land order
Merge this -> uplift `third_party/tt_forge_models` on tt-xla -> add the
`pyramid_flow/pytorch-Vae-single_device-inference` runner entry (written and staged
locally, `EXPECTED_PASSING`). No companion PR and no submodule bump are needed to
merge this one.

### Checklist

### DIFF
--- a/pyramid_flow/pytorch/loader.py
+++ b/pyramid_flow/pytorch/loader.py
@@ -13,9 +13,16 @@ Pyramid Flow has no diffusers integration, so the model code is vendored in
 `src/flux_modules/` (verbatim from upstream, with the `trainer_misc` sequence-
 parallel imports replaced by a local stub).
 
+The CausalVideoVAE has no diffusers integration either, so its decode path is
+vendored in `src/video_vae/` on the same terms, with the context-parallel
+`utils` imports replaced by a local stub. Only the modules decode needs are
+vendored; upstream's training-only loss, LPIPS and discriminator modules are
+left out.
+
 Components:
   - miniFLUX_768p  -> PyramidFluxTransformer DiT (1.97B)
   - ClipTextEncoder -> CLIPTextModel pooled embedding (0.12B)
+  - Vae            -> CausalVideoVAE decode, latents -> pixels (0.23B)
 """
 
 from dataclasses import dataclass
@@ -38,6 +45,8 @@ from .src.utils import (
     load_text_encoder_inputs,
     load_transformer,
     load_transformer_inputs,
+    load_vae_decoder,
+    load_vae_decoder_inputs,
 )
 
 
@@ -53,6 +62,7 @@ class ModelVariant(StrEnum):
 
     MINIFLUX_768P = "miniFLUX_768p"
     CLIP_TEXT_ENCODER = "ClipTextEncoder"
+    VAE = "Vae"
 
 
 class ModelLoader(ForgeModel):
@@ -64,10 +74,11 @@ class ModelLoader(ForgeModel):
 
       - MINIFLUX_768P: the PyramidFluxTransformer DiT, the model's compute core.
       - CLIP_TEXT_ENCODER: the CLIP encoder producing the DiT's `pooled_projections`.
+      - VAE: the CausalVideoVAE decode that turns generated latents into pixels.
 
     Not exposed: the T5-XXL encoder (`text_encoder_2`) compiles and runs on
-    device but reaches only PCC 0.86 against CPU, and the CausalVideoVAE and
-    flow-matching scheduler have not been brought up. Those remain CPU-side.
+    device but reaches only PCC 0.86 against CPU, and the flow-matching
+    scheduler has not been brought up. Those remain CPU-side.
     """
 
     _VARIANTS = {
@@ -76,6 +87,10 @@ class ModelLoader(ForgeModel):
             source=ModelSource.HUGGING_FACE,
         ),
         ModelVariant.CLIP_TEXT_ENCODER: PyramidFlowConfig(
+            pretrained_model_name="rain1011/pyramid-flow-miniflux",
+            source=ModelSource.HUGGING_FACE,
+        ),
+        ModelVariant.VAE: PyramidFlowConfig(
             pretrained_model_name="rain1011/pyramid-flow-miniflux",
             source=ModelSource.HUGGING_FACE,
         ),
@@ -108,12 +123,16 @@ class ModelLoader(ForgeModel):
         dtype = dtype_override if dtype_override is not None else torch.float32
         if self._variant == ModelVariant.CLIP_TEXT_ENCODER:
             return load_text_encoder(dtype)
+        if self._variant == ModelVariant.VAE:
+            return load_vae_decoder(dtype)
         return load_transformer(dtype)
 
     def load_inputs(self, dtype_override=None, **kwargs) -> Any:
         dtype = dtype_override if dtype_override is not None else torch.float32
         if self._variant == ModelVariant.CLIP_TEXT_ENCODER:
             return load_text_encoder_inputs(dtype)
+        if self._variant == ModelVariant.VAE:
+            return load_vae_decoder_inputs(dtype)
         return load_transformer_inputs(dtype)
 
     def unpack_forward_output(self, output: Any) -> torch.Tensor:
--- a/pyramid_flow/pytorch/src/utils.py
+++ b/pyramid_flow/pytorch/src/utils.py
@@ -21,6 +21,8 @@ _DIT_SUBFOLDER = "diffusion_transformer_768p"
 # exposed here - see `load_text_encoder` for why.
 _CLIP_SUBFOLDER = "text_encoder"
 _CLIP_TOKENIZER_SUBFOLDER = "tokenizer"
+# The CausalVideoVAE that decodes miniFLUX latents back to pixels.
+_VAE_SUBFOLDER = "causal_video_vae"
 
 
 # ============================================================================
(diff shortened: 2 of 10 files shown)


ctr-lelanchelianTT:
**Date:** 25/08/2026

**Updates:**

Opened this PR today for the Pyramid Flow CausalVideoVAE decoder, the last component of this model with no TT path — the loader docstring recorded it as un-brought-up until this change. Jira: TEN-1066.

Work completed today on **n150** (`ARCH_NAME=wormhole_b0`), in three steps: vendored the decode-path modules into `src/video_vae/` on the same terms as the DiT's `src/flux_modules/`; added the `Vae` variant with `CausalVaeDecoderWrapper`; validated on device.

Two things I checked rather than assumed, both now recorded in the code:

- Upstream `decode_latent` calls the **chunked** decode, which walks temporal windows in a Python loop carrying causal-conv state and does not trace. This PR compiles the single-shot path instead, and the two are **bit-exact at T=1** (max abs diff 0.0) — so the traceable path is the same math the pipeline runs for one frame, not a substitute.
- A correct decode first measured PCC **0.034**, which looked like a lowering bug. It was not: `torch.randn` consumes its generator differently per dtype, so a seeded bf16 draw is unrelated to the fp32 draw from the same seed (PCC 0.0098 between the two latents) — the two legs were decoding different inputs. `load_inputs` now draws in fp32 and casts.

Validation results: CPU fp32 vs TT bf16 **PCC 0.999956**, MAE 0.005481, max abs diff 0.059273 on golden max abs 2.093321, reproduced three times. Runner id `pyramid_flow/pytorch-Vae-single_device-inference` **PASSED (1 passed in 106.73s)**; the model goes from 12 exposed ids to 18. Conv3d lowers — `prepare_conv3d_weight` is visible in the dumps. CI is green: pre-commit plus all four Cycode checks.

Deliberately kept in **draft** — internal review is required before this goes to customer review.

**Next Steps:**

Internal review. Two choices are the ones I would push back on first if I were reviewing: the ~3.5k lines of vendored `video_vae/` code (limited to the decode path; upstream's training-only loss, LPIPS and discriminator modules are excluded), and validating at `T=1` only. Happy to adjust either. Once internally approved I will move this out of draft, send it for customer review, and move TEN-1066 to Customer Review.

**Challenges / Blockers:**

None technical — the component works and CI is green. The tt-xla runner YAML entry is written but deliberately not raised yet; it needs this PR merged and the `third_party/tt_forge_models` submodule uplifted first (tracked as TEN-1071).

PR #904, opened 2026-08-27
[InternLM2] Sanitize null-char SPM piece so the tokenizer loads

devisettymahidhar608:
## Summary
- `internlm2/pytorch-internlm2-chat-20b` fails to load its tokenizer: `AutoTokenizer.from_pretrained(..., trust_remote_code=True, use_fast=False)` returns the Python `bool` `False`, and the loader then crashes with `'bool' object has no attribute 'apply_chat_template'`.
- Root cause: InternLM2's `tokenizer.model` contains a single NORMAL SentencePiece piece whose surface is a literal null character. `sentencepiece>=0.2.0` rejects it (`RuntimeError: piece must not include null character`), and transformers 4.46.3 swallows that error inside `_from_pretrained`, silently returning `False`.
- Fix: before loading, rewrite that one piece in the cached `tokenizer.model` to a unique non-null placeholder (`<0x00_nul>`). The token id is preserved and the null piece never appears in real text, so tokenization/logits are unchanged. The rewrite is idempotent (skips once the null byte is gone).

### Why not pin `sentencepiece<0.2`?
There is no cp312 wheel for `sentencepiece==0.1.99`, and the sdist build fails in CI (`No module named 'cmake'` / no system `libsentencepiece`), so the pin is not viable.

ci link: https://github.com/tenstorrent/tt-xla/actions/runs/33039896454

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internlm2/pytorch/loader.py
+++ b/internlm2/pytorch/loader.py
@@ -76,6 +76,57 @@ class ModelLoader(ForgeModel):
             framework=Framework.TORCH,
         )
 
+    def _sanitize_spm_null_piece(self):
+        """Make InternLM2's ``tokenizer.model`` loadable under sentencepiece>=0.2.
+
+        The checkpoint's SentencePiece model contains a single NORMAL piece whose
+        surface is a literal null character. sentencepiece>=0.2.0 rejects it with
+        ``RuntimeError: piece must not include null character``; transformers
+        4.46.3 catches that inside ``_from_pretrained`` and silently returns
+        ``False`` from ``AutoTokenizer.from_pretrained`` -- so the loader ends up
+        with ``self.tokenizer = False`` and later hits
+        ``'bool' object has no attribute 'apply_chat_template'``.
+
+        Pinning sentencepiece<0.2 is not viable (no cp312 wheel; the sdist build
+        fails), so instead we rewrite that one piece in the cached model file to a
+        unique non-null placeholder. The token id is preserved, and the null
+        piece never appears in real text, so tokenization/logits are unchanged.
+        The rewrite is idempotent (the null byte is gone after the first pass).
+        """
+        from huggingface_hub import hf_hub_download
+
+        spm_path = hf_hub_download(
+            self._variant_config.pretrained_model_name, "tokenizer.model"
+        )
+        with open(spm_path, "rb") as f:
+            blob = f.read()
+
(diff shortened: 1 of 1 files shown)


sonalibaskaran2499 (approved):

kmabeeTT (changes requested):
The diagnosis and the "can't pin 0.1.99 (no cp312 wheel)" reasoning are both solid, so the approach is right. One thing to fix before merge:

  `_sanitize_spm_null_piece` rewrites `tokenizer.model` **in place inside the shared HF cache** (the `hf_hub_download` path, usually a symlink into `blobs/`). It passes here because the CI cache is writable — but the `open(..., "wb")` will **raise on a read-only or shared/NFS cache**, which some CI and multi-user
  setups use, and it edits a file the loader doesn't own. It's a latent failure I'd rather not bank on the cache always being writable.

  Please sanitize a **local copy** instead:

  1. `snapshot_download(...)` the tokenizer files
  2. `shutil.copy2(os.path.realpath(f), tmp)` — dereferences the blob symlinks so the cache blobs are never touched
  3. sanitize the copy in `tmp`
  4. `AutoTokenizer.from_pretrained(<tmp_dir>, trust_remote_code=True, use_fast=False)`

  Keeps the cache pristine and works regardless of cache permissions. Worth a quick check that `trust_remote_code` still resolves from a local dir on 4.46.3 (it should — reads `auto_map` + imports the staged `.py`).

devisettymahidhar608:
Addressed the cache-mutation concern: `_sanitize_spm_null_piece` no longer writes into the shared HF cache.

The loader now:
1. `snapshot_download(...)`s only the tokenizer files (`allow_patterns`, so the 20B weights are not pulled)
2. `shutil.copy2(os.path.realpath(f), tmp)` into a `TemporaryDirectory` (dereferences blob symlinks; cache is never touched)
3. sanitizes `tokenizer.model` in that copy
4. `AutoTokenizer.from_pretrained(tmp_dir, trust_remote_code=True, use_fast=False)`

Checked locally against transformers 4.46.3: `trust_remote_code` resolves from the staged dir (`auto_map` + the copied `.py`), `apply_chat_template` works, and the cache blob hash/mtime is unchanged.

kmabeeTT (approved):
Thx for change!

